PR #2, opened 2017-03-09
Add lerna bootstrap to npm postinstall

jleft:
`lerna bootstrap` will now be automatically run after `npm install`, simplifying the setup process and removing the need for the user to install Lerna globally.


BenLambertNcl (approved):

ColinEberhardt:
Nice idea :-)

PR #7, opened 2017-03-10
OpenFin Notifications

BenLambertNcl:
Basic implementation of the OpenFin notifications.

The OpenFin API requires a html template to use as its notification. I have included a very simple html template that resembles the default electron notification. OpenFin does not support native notifications.

OpenFin does not allow preload scripts like electron. Because of this, `npm posinstall` copies the API into the `src` directory, which is accessed via a script tag. I also included a blank `index.js` file in the specification project to avoid any 404 errors from electron/browsers.


packages/api-openfin-demo/package.json
```diff
@@ -0,0 +1,24 @@
...
+ "main": "",
+ "scripts": {
+   "serve": "live-server --port=5000 --no-browser",
+   "launch": "openfin --launch --config app.json",
+   "postinstall": "copyfiles -f node_modules/ssf-desktop-api-specification/src/** src && copyfiles -f node_modules/ssf-desktop-api-openfin/** src"
+ },
+ "author": "",
+ "license": "Apache-2.0",
+ "dependencies": {
+   "bootstrap": "^3.3.7",
```
jleft: I don't think this is used - remove?

packages/api-openfin-demo/package.json
```diff
@@ -0,0 +1,24 @@
...
+   "jquery": "^3.1.1",
```
jleft: I don't think this is used - remove?

packages/api-openfin-demo/app.json
```diff
@@ -0,0 +1,19 @@
...
+    "runtime": {
+        "arguments": "",
+        "version": "stable"
+    },
+    "shortcut": {
+        "company": "",
+        "description": "",
+        "name": "SSF API Demo Openfin"
+    }
+}
```
jleft: Nitpick: add newline at the end of the file

packages/api-openfin/README.md
```diff
@@ -0,0 +1,13 @@
+# Symphony API for Openfin
```
jleft: _Openfin_ => _OpenFin_

packages/api-openfin/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This project provides an implementation of the Symphony Desktop Wrapper API Specification for Openfin
```
jleft: _Openfin_ => _OpenFin_

packages/api-openfin/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Usage
+
+Add this package to your openfin project:
```
jleft: _openfin_ => _OpenFin_

packages/api-openfin/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+```
+npm install ssf-desktop-api-openfin --save
+```
+
+You need to include this code into the project by loading the javascript file directly (as Openfin does not support preloading).
```
jleft: _javascript_ => _JavaScript_
_Openfin_ => _OpenFin_

packages/api-specification/src/index.js
```diff
@@ -0,0 +1 @@
+// Intentionally blank, overridden by openfin to access API
```
jleft: _openfin_ => _OpenFin_

packages/api-openfin-demo/README.md
```diff
@@ -0,0 +1,11 @@
+# Symphony API for Openfin Demo
```
jleft: Openfin has the _f_ capitalised. Please can _Openfin_ be changed to _OpenFin_?

packages/api-openfin-demo/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This project provides a demonstration of the Symphony Desktop Wrapper API within Openfin.
```
jleft: _Openfin_ => _OpenFin_

packages/api-openfin/index.js
```diff
@@ -0,0 +1,13 @@
+/* globals fin  */
```
jleft: I think this could be required in multiple files in the future - can this be added to a `.eslintrc.json` file at the root of this package,  `packages/api-openfin/`?

```json
{
    "globals": {
        "fin": true
    }
}
```

packages/api-openfin/index.js
```diff
@@ -0,0 +1,13 @@
...
+window.Notification = function(title, options) {
+  const message = {
+    title: title,
+    options: options
```
jleft: Will these options do anything?

I.e. if I look in our API specification, the options include `dir`, `lang` etc. If they don't do anything then I think we can omit them from the object passed to OpenFin and only pass the message, but keep `options` in our API and we can add support for dir`, `lang` etc. later.

packages/api-openfin/README.md
ColinEberhardt: JavaScript
OpenFin

Yes, I really am that pedantic!

packages/api-openfin/notification.html
```diff
@@ -0,0 +1,11 @@
+<h5 id="message">Message</h5>
```
jleft: Assuming that this page needs to be valid HTML, worth updating to something like this?

```html
<!DOCTYPE html>
<meta charset="utf-8">
<title></title>
<style>
  html {
    border: solid 2px black;
  }
</style>
<h5 id="message"></h5>
<script>
  function onNotificationMessage(message) {
    document.getElementById('message').innerText = message.title;
  }
</script>
```

Might be worth keeping the `h5` empty too, without the text `Message`?

```html
<h5 id="message"></h5>
```

ColinEberhardt (approved):
Apart from a very pedantic comment about capitalisation, looks good to me :+1:

packages/api-specification/src/index.js
jleft: Will this also be an issue in the Electron version?

jleft (changes requested):
A couple of questions and some very minor comments, but overall looking really good to me!

packages/api-openfin/README.md
jleft: Beat me to it! :smile:

packages/api-specification/src/index.js
BenLambertNcl: The electron version will load this empty file, otherwise it omits a 404 error.

jleft: :+1:

packages/api-openfin/notification.html
BenLambertNcl: OpenFin notifications are html pages in their own right, and this is just injected inside the body of that page. I can still change it to be like this (but without the `<!DOCTYPE html>`) if you prefer?

jleft: >OpenFin notifications are html pages in their own right, and this is just injected inside the body of that page.

Are you sure that this is just injected into the body? From the docs  and examples it looks like it's the entire HTML page.

BenLambertNcl: image

This is the output from the notification box, with extra tags for html, body and head that were not included in the template (This is with the changes you suggested, bar the doctype).

ColinEberhardt: AFAIK that's just the way Chrome handles minimal HTML5 documents. It adds in the `html`, `head`, `body` elements in the correct locations.

jleft: Ah, I see - when using dev tools, you're not seeing the original HTML source of the page which has been loaded - it's showing the DOM, the browsers interpretation of the HTML.

The browser will try to add missing elements, correct invalid markup etc. Therefore the original HTML can differ from the DOM shown in the dev tools. I don't OpenFin is injecting the template into the body here, I think it's just the browser.

It's still very minor, as the browser can correct it, but I think it's worth ensuring the HTML is valid, with doctype et. al.

BenLambertNcl: That makes sense. :+1:

jleft (approved):

PR #8, opened 2017-03-13
Notification body

BenLambertNcl:
Added body to notifications. I altered the styles for the OpenFin template slightly to match the electron notification more closely (As it used default HTML padding/margins).

### DIFF
--- a/packages/api-openfin/index.js
+++ b/packages/api-openfin/index.js
@@ -1,4 +1,8 @@
 window.Notification = function(title, options) {
+  if (!options) {
+    options = {};
+  }
+
   const message = {
     title: title,
     text: options.body
--- a/packages/api-openfin/index.js
+++ b/packages/api-openfin/index.js
@@ -1,6 +1,7 @@
 window.Notification = function(title, options) {
   const message = {
-    title: title
+    title: title,
+    text: options.body
   };
 
   // eslint-disable-next-line no-new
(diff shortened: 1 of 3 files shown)


packages/api-openfin/notification.html
```diff
@@ -11,13 +11,20 @@
       right: 0px;
       bottom: 0px;
     }
+
+    h4, p {
```
jleft: Nitpick: prefer selectors on newlines for readability:

```css
h4,
p {
```

packages/api-openfin/notification.html
```diff
@@ -11,13 +11,20 @@
...
+      padding: 0px;
+      margin: 0px;
```
jleft: Even more pedantic... `0px` is consistent with the styles above, which is good, but the `px` is unnecessary; something to consider removing. I.e. `padding: 0;` etc. is sufficient and a little less cluttered IMO.

packages/api-openfin/index.js
```diff
@@ -1,6 +1,11 @@
 window.Notification = function(title, options) {
```
jleft, line 1: Could considering using default param here: `options = {}`

jleft (commented):
A few minor comments, but looking good to me

packages/api-openfin/index.js
jleft: Actually, this would change the behaviour (slightly) and be inconsistent with Electron implementation - maybe best left as-is :+1:

jleft (approved):

PR #11, opened 2017-03-13
Remove src directory in Electron and OpenFin demos before copying files

jleft:


packages/api-openfin-demo/package.json
```diff
@@ -1,22 +1,23 @@
 {
- "name": "ssf-desktop-api-openfin-demo",
```
BenLambertNcl, line 2: Have the line endings changed in this file? Seems odd the whole file changed

jleft: It's the indentation - it's updated from 1 space, to 2 spaces. Happened automatically when I ran `npm i`

BenLambertNcl (approved):

PR #13, opened 2017-03-14
Independent window demos

BenLambertNcl:
Fully independent windows in both OpenFin and Electron use the HTML5 API. This is a demo to showcase that functionality. 

**Note:** Reopening a window with the same name parameter does not work in Electron when not using `window.open`, but this method has the side effect of creating a child window (i.e. the new window closes when the parent closes). It also does not work in OpenFin, and is [inconsistent across browsers](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) (See note about strWindowName). For this reason, we decided that having a fully independent window was more important, but it should be possible to add this functionality in manually at a later date.

### DIFF
--- a/packages/api-electron/index.js
+++ b/packages/api-electron/index.js
@@ -1,4 +1,7 @@
 const ipc = require('electron').ipcMain;
+const { BrowserWindow } = require('electron');
+
+let windows = [];
 
 module.exports = () => {
   const eNotify = require('electron-notify');
@@ -12,5 +15,12 @@ module.exports = () => {
       title: msg.title,
       text: msg.options.body
     });
+  });
+
+  ipc.on('ssf-new-window', (e, msg) => {
+    const newWin = new BrowserWindow();
+    newWin.loadURL(msg.url);
+    windows.push(newWin);
+    e.returnValue = newWin;
   });
 };
--- a/packages/api-electron/index.js
+++ b/packages/api-electron/index.js
@@ -1,7 +1,9 @@
-const ipc = require('electron').ipcMain;
-const { BrowserWindow } = require('electron');
+const {
+  BrowserWindow,
+  ipcMain: ipc
+} = require('electron');
 
-let windows = [];
+const windows = [];
 
 module.exports = () => {
   const eNotify = require('electron-notify');
@@ -18,9 +20,15 @@ module.exports = () => {
   });
 
   ipc.on('ssf-new-window', (e, msg) => {
-    const newWin = new BrowserWindow();
-    newWin.loadURL(msg.url);
-    windows.push(newWin);
-    e.returnValue = newWin;
+    const newWindow = new BrowserWindow();
+    newWindow.loadURL(msg.url);
+    windows.push(newWindow);
+    newWindow.on('close', () => {
+      const index = windows.indexOf(newWindow);
+      if (index >= 0) {
+        windows.splice(index, 1);
+      }
+    });
+    e.returnValue = newWindow;
   });
 };
--- a/packages/api-specification/src/new-window-api.html
+++ b/packages/api-specification/src/new-window-api.html
@@ -24,6 +24,10 @@
     <label>URL</label>
     <input type="text" class="form-control" placeholder="url" id="url">
   </div>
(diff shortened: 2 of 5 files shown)


packages/api-specification/src/new-window-api.html
```diff
@@ -0,0 +1,67 @@
...
+  </nav>
+  <h1>Symphony Desktop Wrapper API Demo</h1>
+</header>
+
+<h2>New Window API</h2>
+
+<h3>Try me</h3>
+
+<form class="form-inline">
+  <div class="form-group">
```
jleft, line 23: How about adding the _name_ as a input too?

packages/api-specification/src/new-window-api.html
```diff
@@ -0,0 +1,67 @@
...
+
+<p>The window API is used to create a new browser window. <a href='https://developer.mozilla.org/en-US/docs/Web/API/Window/open'>HTML5 Window API</a></p>
+
+<h3>Method Signature</h3>
+
+<pre>
+window.open(url, name, features)
+</pre>
+
+<p>Creates a new window.</p>
```
jleft: Maybe add that if a window already exists with the same `name` then that window will be reused, rather than a new window being created?

packages/api-specification/src/new-window-api.html
```diff
@@ -0,0 +1,67 @@
...
+  <dd>Name of the new window. Should not contain whitespace</dd>
+  <dt>features (optional)</dt>
+  <dd>DOMString to set features of the new window</dd>
+</dl>
+
+<h4>Return Value</h4>
+
+<dl>
+  <dt>window</dt>
+  <dd>A reference to the new browser window.</dd>
```
jleft: Maybe just _window_ not _browser window_?

packages/api-electron/index.js
```diff
@@ -1,4 +1,7 @@
 const ipc = require('electron').ipcMain;
+const { BrowserWindow } = require('electron');
```
jleft: Could combine this to a single require (I think)
```js
const {
  BrowserWindow,
  ipcMain: ipc
} = require('electron');
```

packages/api-electron/index.js
```diff
@@ -1,4 +1,7 @@
...
+
+let windows = [];
```
jleft: Could be a `const`, as I can't see it being reassigned.

packages/api-electron/index.js
```diff
@@ -13,4 +16,11 @@ module.exports = () => {
       text: msg.options.body
     });
   });
+
+  ipc.on('ssf-new-window', (e, msg) => {
+    const newWin = new BrowserWindow();
```
jleft: I'd probably be a bit more verbose: `newWin` -> `newWindow`

packages/api-electron/index.js
```diff
@@ -13,4 +16,11 @@ module.exports = () => {
...
+    newWin.loadURL(msg.url);
+    windows.push(newWin);
```
jleft: Is this a potential memory leak, if the window isn't removed from `windows` when it's closed?

Could we remove the closed window reference from `windows` when its `close` or `closed` event is fired?

packages/api-electron/index.js
```diff
@@ -13,4 +18,14 @@ module.exports = () => {
       text: msg.options.body
     });
   });
+
+  ipc.on('ssf-new-window', (e, msg) => {
+    const newWindow = new BrowserWindow();
+    newWindow.loadURL(msg.url);
+    windows.push(newWindow);
+    newWindow.on('close', () => {
+      windows.splice(windows.indexOf(newWindow), 1);
```
jleft: I think we should only `splice` if `indexOf` returns `> -1` - I know it's an unlikely that it will return `-1` but I think it's worth being defensive here.

jleft (commented):
Just one more comment, but looking good :+1:

packages/api-electron/index.js
BenLambertNcl: Yeah good point :)

jleft: :+1:

PR #17, opened 2017-03-15
Refactor of method signature for creating new windows

BenLambertNcl:
Change from window.open => ssf.window for opening new windows.

Reason for this change was because creating windows in OpenFin (using `fin.desktop.Window`) involves calling window.open, giving us an endless loop if we override window.open.

To make this change, I needed to add a browser API which is done by including the JavaScript files (the same as OpenFin).


packages/api-browser-demo/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "ssf-desktop-api-browser-demo",
+  "version": "0.0.1",
+  "description": "",
+  "main": "",
+  "scripts": {
+    "serve": "live-server --port=5000 src/",
+    "start": "npm run serve",
```
jleft: Could probably remove `serve` and just have `start` as `live-server --port=5000 src/`

Not sure if it's worth specifying the port - could remove?

packages/api-browser-demo/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "postinstall": "rimraf src && copyfiles -f node_modules/ssf-desktop-api-specification/src/** src && copyfiles -f node_modules/ssf-desktop-api-browser/** src"
+  },
+  "author": "",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "openfin-cli": "^1.1.5",
```
jleft: Looks to be unused - remove.

packages/api-specification/src/new-window-api.html
```diff
@@ -38,7 +38,7 @@
 <h3>Method Signature</h3>
 
 <pre>
-window.open(url, name, features)
+ssf.Window(url, name, features)
```
jleft: I think we'll need to consider whether the API will follow the factory pattern (used here), or the constructor pattern. For example:

```js
// Factory
var newWindow = ssf.window();

// Constructor
var newWindow = new ssf.Window();
```

From what I can see on the Wiki, there seems to be a preference for the constructor pattern, but I'm happy to use the factory pattern, get some feedback, then change if required.

However, if we are going to stick with the factory pattern for now, I think the casing should be changed. PascalCase for classes and constructors and camelCase for everything else is a common convention, so `ssf.Window()` => `ssf.window()`

packages/api-browser/window.js
```diff
@@ -0,0 +1,5 @@
+if (!window.ssf) {
+  window.ssf = {};
+}
+
+window.ssf.Window = (url, name, features) => window.open(url, name, features);
```
jleft: Could use rest params here:

```js
(...args) => window.open(...args)
```

packages/api-openfin/window.js
```diff
@@ -0,0 +1,5 @@
+if (!window.ssf) {
+  window.ssf = {};
+}
+
+window.ssf.Window = (url, name, features) => window.open(url, name, features);
```
jleft: Could use rest params here:

```js
(...args) => window.open(...args)
```

jleft (commented):
:+1: only a few style comments

packages/api-browser-demo/package.json
BenLambertNcl: It's used in `launch`, unless I'm mistaken?

packages/api-specification/src/new-window-api.html
BenLambertNcl: Yeah sounds like a good idea. Won't be a big change if the feedback is to use the constructor pattern :)

jleft (approved):
LGTM - thanks!

PR #20, opened 2017-03-16
Child windows

BenLambertNcl:
Added option to create child windows by passing `child=yes` as a feature to the `ssf.window` function. 

Child windows are closed at the same time as the parent, whereas non child windows are created in a new process and continue to live after the parent has been closed. Browsers do not support this functionality and will just ignore the `child=yes` feature.


packages/api-electron/index.js
```diff
@@ -20,15 +20,37 @@ module.exports = () => {
   });
 
   ipc.on('ssf-new-window', (e, msg) => {
-    const newWindow = new BrowserWindow();
+    let options = {};
```
jleft: Can't see this being reassigned, so: `const`

packages/api-electron/index.js
```diff
@@ -20,15 +20,37 @@ module.exports = () => {
...
+
     e.returnValue = newWindow;
+    windows.push(newWindow);
+  });
+};
+
+const parseFeaturesString = (features) => {
+  const featureObject = {};
+
+  features.split(/,\s*/).forEach((feature) => {
```
jleft: I see this matches whitespace after the comma, is it valid to have whitespace in the string?

packages/api-openfin/window.js
```diff
@@ -2,4 +2,48 @@ if (!window.ssf) {
   window.ssf = {};
 }
 
-window.ssf.window = (...args) => window.open(...args);
+window.ssf.window = function(url, name, features) {
+  let newWindow;
+  const featureObject = parseFeaturesString(features);
+  const handleError = (error) => console.error('Error creating window: ' + error);
```
jleft, line 8: I think we'll need to consider some way of allowing the user to handle errors themselves - allowing them to pass their own callback to handle the error, for example.

packages/api-electron/index.js
```diff
@@ -20,15 +20,37 @@ module.exports = () => {
...
+    let [key, value] = feature.split(/\s*=/);
+
+    // interpret the value as a boolean, if possible
+    value = (value === 'yes' || value === '1') ? true : (value === 'no' || value === '0') ? false : value;
```
jleft: Nitpick: I'm not a big fan of nested ternary operators - I think they can be difficult to understand - maybe consider using `if`/`else`

packages/api-specification/src/new-window-api.html
```diff
@@ -60,7 +64,12 @@
     <dt>name</dt>
     <dd>Name of the new window. Should not contain whitespace</dd>
     <dt>features (optional)</dt>
-    <dd>DOMString to set features of the new window</dd>
```
jleft, line 63: I know one of the principles is to follow the HTML5 API, but I think an object, rather than a string, would be much nicer for the consumer.

For example,  Electron's [BrowserWindow](https://github.com/electron/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions) has more than 40 options (nearly 70 if we include the options in `webPreferences`), I think building/maintaining a string with more than a handful of options could be pretty painful for the user.

```js
ssf.window(url, windowName, `child=${isChild},height=${height},width=${width},x=${x},y=${y}...`);
```

I think this is more of a discussion point, than something to fix, at the moment - but maybe having an API like this would work well:

```js
// options probably required, successCallback and errorCallback optional
new ssf.Window(options, successCallback, errorCallback);

// example usage
new ssf.Window({
  child: isChild,
  height,
  width,
  x,
  y
});

```

What do you think @BenLambertNcl? (cc @ColinEberhardt)

jleft (commented):
Looks good :+1: - just a couple of small points/questions and a comment for more general discussion (probably not something to fix now, but I thought it was worth raising).

packages/api-specification/src/new-window-api.html
BenLambertNcl: I 100% agree, but I thought I would try to stick to the HTML5 API as much as possible with this PR.

packages/api-electron/index.js
BenLambertNcl: This is the same regex electron uses when parsing the feature list for window.open, but MDN states that the string must not contain any white space. Ill remove it to keep it consistent with the HTML5 API

packages/api-openfin/window.js
jleft: (Not necessarily something to do in the PR.)

BenLambertNcl: Yeah that sounds like a good idea. Maybe if we change the features string to an object, we can add this at the same time

packages/api-openfin/window.js
```diff
@@ -2,4 +2,48 @@ if (!window.ssf) {
...
+    }, () => app.run(), handleError);
+
+    // Need to return the window object, not the application
+    newWindow = app.getWindow();
+  }
+
+  return newWindow;
+};
+
+const parseFeaturesString = (features) => {
```
jleft, line 36: Thanks for updating this function in the Electron API - please can these changes also be replicated here?

jleft (changes requested):

packages/api-openfin/window.js
BenLambertNcl: Oops, forgot about this one :+1:

jleft (approved):

PR #21, opened 2017-03-16
Add ScreenSnippet API

jleft:


packages/api-specification/src/html2canvas.js
```diff
@@ -0,0 +1 @@
+// Intentionally blank, overridden to in platform demos
```
BenLambertNcl: Remove `to` here? It reads wrong to me.

packages/api-openfin/screen-snippet.js
```diff
@@ -0,0 +1,14 @@
+/* globals createImageBitmap, html2canvas */
```
BenLambertNcl: Possibly worth adding `createImageBitmap` into `eslintrc.json` as its declared as global in more than 1 file?

BenLambertNcl (commented):
LGTM

BenLambertNcl (approved):

PR #23, opened 2017-03-17
Hidden main window

BenLambertNcl:
Create an initial window that is hidden, where we can create new windows via the windowing API. This is to create a common bootstrapping process between Electron and OpenFin. However, the browser has no concept of this so it will just startup normally.

Electron and OpenFin will both load into `hiddenWindow.html` where we can load `index.html` through the windowing API (e.g. `ssf.window('http://localhost:5000/index.html', ...)`).

_Note:_ This implementation required serving the files for electron instead of loading the files from the file system. This was necessary as chrome's sandbox disables reading from the file system once the render process starts.


packages/api-electron/README.md
```diff
@@ -20,23 +20,15 @@ This module needs to register IPC handlers once the app `ready` event has fired.
...
-  ssfElectron();
+  ssfElectron(url);
   // ... other stuff!
 }
 
 app.on('ready', createWindow)
 ```
 
-### Renderer process integration
+where `url` is the url to load in the hidden window.
```
jleft: Nitpick: url => URL (not the code, but the description)

packages/api-electron/README.md
```diff
@@ -20,23 +20,15 @@ This module needs to register IPC handlers once the app `ready` event has fired.
...
-```
-win = new BrowserWindow({
-  // ...
-  webPreferences: {
-    // ...
-    preload: path.join(__dirname, 'node_modules/ssf-desktop-api-electron/preload.js')
-  }
-})
-```
+The call to ssfElectron will create a BrowserWindow that will not be shown, and will load the url passed into `ssfElectron()`. From here, use `ssf.window()` to create windows.
```
jleft: Maybe just _browser window_, instead of _BrowserWindow_? The latter feels like an internal implementation detail, that I don't think the user needs on know about.

packages/api-specification/src/hiddenWindow.html
```diff
@@ -0,0 +1,7 @@
+<!doctype html>
```
jleft: Feels a little odd to me that this is in the api-specification and not in the demo packages - I guess the benefit having it here is that it doesn't need to be duplicated?

packages/api-specification/src/hiddenWindow.html
```diff
@@ -0,0 +1,7 @@
...
+<body>
+  <script src="window.js"></script>
+  <script>
+    ssf.window('http://localhost:5000/index.html', 'name', 'child=yes');
```
jleft: Could give a more descriptive name? `'hiddenWindow'`?

packages/api-specification/src/hiddenWindow.html
jleft: I wonder if _mainWindow_, _masterWindow_, _hiddenMainWindow_ or something like that would be a better name? To indicate to the user the this isn't just a normal window which is hidden, but something different?

jleft (commented):
Looks good - I like this :+1:

packages/api-specification/src/hiddenWindow.html
BenLambertNcl: Yeah it was in api-specification to avoid the duplication, but I could move it if you think that's better? I think _hiddenMainWindow_ is probably my preference from those.

packages/api-specification/src/hiddenWindow.html
BenLambertNcl: As this loads the main application, would something like _mainWindow_ be better?

packages/api-specification/src/hiddenWindow.html
jleft: I think it's okay for now - at some point I expect we'll need to split up the demo/code from the specification to make it a bit easier to maintain.

> I think hiddenMainWindow is probably my preference from those.

:+1:

packages/api-specification/src/hiddenWindow.html
jleft: Sounds good to me :+1:

jleft (approved):

ColinEberhardt:
Nice idea :-)

PR #27, opened 2017-03-20
Update Electron Screen Snippet API to use capturePage

jleft:


ColinEberhardt:
Does this mean `html2canvas` is no longer required as a dependency?

BenLambertNcl (approved):

jleft:
> Does this mean `html2canvas` is no longer required as a dependency?

Yes, good point - I've removed it.

PR #29, opened 2017-03-21
Add App API with ready method

jleft:


packages/api-specification/src/new-window-api-demo.js
```diff
@@ -4,6 +4,8 @@ newWindowButton.onclick = function() {
   var url = document.getElementById('url').value;
   var windowName = document.getElementById('name').value;
   var isChild = document.getElementById('child').checked ? 'yes' : 'no';
-  // eslint-disable-next-line no-new
-  new ssf.Window(url, windowName, 'child=' + isChild);
+  ssf.app.ready().then(() => {
```
BenLambertNcl: Possibly wrap on click inside this callback instead of the other way around.

BenLambertNcl (commented):
LGTM

packages/api-specification/src/new-window-api-demo.js
jleft: Good, point, I've moved the `ready` call to outside of the button click handler

BenLambertNcl (approved):

PR #30, opened 2017-03-22
Message service

BenLambertNcl:
The messaging API allows the sending of JavaScript objects from window to window for Electron and OpenFin.

### DIFF
--- a/packages/api-electron/index.js
+++ b/packages/api-electron/index.js
@@ -55,6 +55,28 @@ module.exports = (url) => {
     });
   });
 
+  ipc.on('message-service', (e, msg) => {
+    const windowId = parseInt(msg.windowId, 10);
+
+    if (isNaN(windowId)) {
+      return;
+    }
+
+    const destinationWindow = BrowserWindow.fromId(windowId);
+
+    if (!destinationWindow) {
+      return;
+    }
+
+    // Need to send to topic and * in case the user has subscribed to the wildcard
+    destinationWindow.webContents.send(`message-service-${msg.topic}`, msg.message, msg.fromId);
+    destinationWindow.webContents.send(`message-service-*`, msg.message, msg.fromId);
+  });
+
+  ipc.on('get-window-id', (e) => {
+    e.returnValue = e.sender.id;
+  });
+
   createInitialHiddenWindow(url);
 };
 
--- a/packages/api-electron/index.js
+++ b/packages/api-electron/index.js
@@ -55,7 +55,7 @@ module.exports = (url) => {
     });
   });
 
-  ipc.on('message-service', (e, msg) => {
+  ipc.on('ssf-send-message', (e, msg) => {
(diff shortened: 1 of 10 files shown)


packages/api-electron/index.js
```diff
@@ -55,6 +55,28 @@ module.exports = (url) => {
     });
   });
 
+  ipc.on('message-service', (e, msg) => {
```
jleft: Should probably prefix all of the events with `ssf-`

Maybe describe what the event is with the event name, like: `ssf-send-message`?

packages/api-electron/preload.js
```diff
@@ -23,6 +23,10 @@ class Window {
       features
     });
   }
+
+  static getCurrentWindowId() {
+    return ipc.sendSync('get-window-id');
```
jleft: Seems odd that we would have to use the main process to get the current window's id - is there not a way to get the window id from the renderer process? Or is this a limitation because of the sandbox?

packages/api-electron/preload.js
```diff
@@ -41,3 +45,26 @@ class ScreenSnippet {
 window.ssf.ScreenSnippet = ScreenSnippet;
 
 window.ssf.app.ready = () => Promise.resolve();
+
+class MessageService {
+  static send(windowId, topic, message) {
+    const fromId = window.ssf.Window.getCurrentWindowId();
```
jleft: Is this needed here, or can the `fromId` be extracted in the main process from `event.sender`?

packages/api-openfin/message.js
```diff
@@ -0,0 +1,28 @@
+if (!window.ssf) {
+  window.ssf = {};
+}
+
+class MessageService {
+  // Window ID should be in the form 'application-uuid:window-name'
+  static send(windowId, topic, message) {
+    const splitId = windowId.split(':');
```
jleft: Could do something like this here, if OpenFin supports this destructuring syntax:

```js
const  [appId, windowName] = windowId.split(':');
```

packages/api-electron/preload.js
BenLambertNcl: Yeah. With the sandbox enabled, there's no way to get the window id from the render process. The recommended way is to use the remote electron module to get the current browser window instance, but this remote module is disabled when the window is sandboxed

packages/api-openfin/message.js
ColinEberhardt: I'm going to make a note of this little trick, I use object destructuring all the time, but forget about arrays!

ColinEberhardt:
Looking really good, I'd happy to see this merged in its current state.

One quick question - Is there a way to send a message to the main 'hidden' window?

ColinEberhardt (approved):

BenLambertNcl:
Sending a message to the hidden window will work the same, as long as the window id is known (this could be sent from the hidden window to its child when it loads, for example).

PR #33, opened 2017-03-22
Change demos to use different ports

### LINKED ISSUE #32: Allow OpenFin & Electron demos to be run at the same time
Both use port 5000, would be good to have a different port for each. Although this is complicated by having them both share `hiddenMainWindow.html`.

BenLambertNcl:
Run all three demos on different ports so they can be run at the same time. Uses `location.port` when creating new windows.

Closes #32


jleft (commented):
LGTM, but it looks like the messaging API demo code in api-specification has a hard-coded URL with port 5000 - should this be updated to work across all platforms?

BenLambertNcl:
Good spot. I think we can use `location.port` in the JavaScript for now, but maybe it would be best to inject this somehow once we have a build stage.

jleft (approved):

PR #35, opened 2017-03-23
Window refactor

BenLambertNcl:
Change the window constructor to accept an object instead of a string. This makes the API much easier to use as well as being closer to the Electron/OpenFin implementations. The browser still uses a string so the object need to be converted back into a string in that API.

Some of the Window API documentation has been updated to include this change, as well as improve the consistency with the other documentation.


packages/api-browser/window.js
```diff
@@ -3,9 +3,26 @@ if (!window.ssf) {
...
   }
 }
 
+const objectToFeaturesString = (features) => {
+  let featuresString = '';
+
+  Object.keys(features).forEach((key, index) => {
+    // Need to convert booleans to yes/no
+    if (features[key] === true) {
+      features[key] = 'yes';
```
jleft: I don't think that we should mutate parameters, as the side effects can lead to confusing behaviour. Maybe we could use `map`, instead of `forEach` to create an array of strings (one string for each feature), then use `arr.join(',')` to create the comma separate list as a string?

jleft (changes requested):
Looks much better than before :+1: Just one comment

packages/api-specification/src/index.html
```diff
@@ -20,7 +20,7 @@
     <ul>
       <li><a href="app-api.html">App</a></li>
       <li><a href="messaging-api.html">Messaging</a></li>
-      <li><a href="new-window-api.html">New Window</a></li>
+      <li><a href="window-api.html">New Window</a></li>
```
jleft: `New Window` => `Window`

jleft (changes requested):
One more comment, then should be good to merge :+1:

packages/api-specification/src/index.html
BenLambertNcl: Oops :+1:

jleft (approved):

PR #37, opened 2017-03-24
Add Browser window messaging

BenLambertNcl:
Browser window messaging is more limited than in Electron or OpenFin due to the lack of main process. The current implementation uses `window.postMessage()`, but this requires a reference to the window you wish to send the message. Therefore, only windows that have been created by `window.open` can be sent messages (as well as the window which opened the current window via `window.opener`).

To keep the browser API in line with the other APIs, new windows are stored with their name in a global object, which can be referenced via an id (the window name, or 'parent' for the opener window).

### DIFF
--- a/packages/api-browser/message.js
+++ b/packages/api-browser/message.js
@@ -1,0 +1,29 @@
+if (!window.ssf) {
+  window.ssf = {};
+}
+
+class MessageService {
+  static send(windowId, topic, message) {
+    const win = window.accessableWindows.find((w) => w.id.toString() === windowId);
+    const senderId = window.ssf.Window.getCurrentWindowId();
+    if (win) {
+      win.window.postMessage({
+        senderId,
+        topic,
+        message
+      }, '*');
+    }
+  }
+
+  static subscribe(windowId, topic, listener) {
+    const receiveMessage = (event) => {
+      if ((windowId === '*' || windowId === event.data.senderId) && topic === event.data.topic) {
+        listener(event.data.message, event.data.senderId);
+      }
+    };
+
+    window.addEventListener('message', receiveMessage, false);
+  }
+}
+
+window.ssf.MessageService = MessageService;
--- a/packages/api-browser/message.js
+++ b/packages/api-browser/message.js
@@ -4,7 +4,7 @@ if (!window.ssf) {
 
 class MessageService {
   static send(windowId, topic, message) {
-    const win = window.accessableWindows.find((w) => w.id.toString() === windowId);
+    const win = window.accessibleWindows[windowId];
     const senderId = window.ssf.Window.getCurrentWindowId();
     if (win) {
       win.window.postMessage({
--- a/packages/api-browser/window.js
+++ b/packages/api-browser/window.js
@@ -2,15 +2,14 @@ if (!window.ssf) {
   window.ssf = {};
 }
 
-window.accessableWindows = [];
-
 class Window {
   constructor(url, name, features) {
     const win = window.open(url, name, objectToFeaturesString(features));
-    window.accessableWindows.push({
-      id: name,
-      window: win
-    });
+    win.onclose = () => {
+      window.accessibleWindows[win.name] = null;
+    };
+
+    window.accessibleWindows[name] = win;
   }
 
   static getCurrentWindowId() {
@@ -32,15 +31,5 @@ const objectToFeaturesString = (features) => {
     return `${key}=${value}`;
   }).join(',');
 };
-
-// if we have an opener, we are not the parent so we need to add it as a window
-if (window.opener) {
-  window.accessableWindows.push({
-    id: 'parent',
-    window: window.opener
-  });
-} else {
-  window.name = 'parent';
-}
 
 window.ssf.Window = Window;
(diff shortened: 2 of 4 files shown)


packages/api-browser/message.js
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  static subscribe(windowId, topic, listener) {
+    const receiveMessage = (event) => {
+      if ((windowId === '*' || windowId === event.data.senderId) && topic === event.data.topic) {
+        listener(event.data.message, event.data.senderId);
+      }
+    };
+
+    window.addEventListener('message', receiveMessage, false);
```
jleft, line 25: In the future, I think it's worth adding an `unsubscribe` method,  as it's useful for the user, but also so we can clean up these event listeners when a window is closed - something which we should consider doing automatically anyway.

packages/api-browser/window.js
```diff
@@ -2,10 +2,20 @@ if (!window.ssf) {
   window.ssf = {};
 }
 
+window.accessableWindows = [];
```
jleft: access*a*ble => access*i*ble

If names are unique, then I think an object might be better here - mapping name to window object. Accessing should be easier and faster, as it would just be `accessibleWindows[name] ` (not `accessibleWindows.find(...)`) and we won't need to create a new object for each window.

However, if order is important and the array is needed, then I don't think the `id` would be needed on the new object pushed into the array, as it's already on the `name` property of the window - could just push the window directly into the array?

If possible, I think we should avoid polluting the global namespace like this. Perhaps, for now, `accessibleWindows` could be a `const` and there could be `static` methods on `Window` to allow this to be get/set?

packages/api-browser/window.js
```diff
@@ -2,10 +2,20 @@ if (!window.ssf) {
...
+
 class Window {
   constructor(url, name, features) {
-    window.open(url, name, objectToFeaturesString(features));
+    const win = window.open(url, name, objectToFeaturesString(features));
+    window.accessableWindows.push({
```
jleft: What happens if the window closes? Feels like it should be removed from here, as it may lead to a memory leak and it would no longer be accessible. Same applies for adding the parent window, below.

packages/api-browser/window.js
```diff
@@ -23,4 +33,14 @@ const objectToFeaturesString = (features) => {
   }).join(',');
 };
 
+// if we have an opener, we are not the parent so we need to add it as a window
+if (window.opener) {
```
jleft: I'm not sure about this living here, feels like it's part to the initialisation/bootstrapping process; probably fine for now, but I think we should consider moving this in future.

jleft (changes requested):

packages/api-browser/window.js
BenLambertNcl: I could move this into `app.js` to be part of the bootstrapping code? Think that should still be simple enough to do in this PR.

jleft: Sure, sounds good to me :+1:

packages/api-browser/window.js
BenLambertNcl: Yeah ill add onclose functions to remove the window

packages/api-browser/window.js
jleft: > If possible, I think we should avoid polluting the global namespace like this. Perhaps, for now, `accessibleWindows` could be a `const` and there could be `static` methods on `Window` to allow this to be get/set?

For this - I get that this would be easier to do this if this code is in the same file or when the ES6 modules are implemented - so let me know if you'd like to discuss further :)

packages/api-browser/message.js
BenLambertNcl: Yeah ill add that as tech debt for when I add to the message service functionality later :+1:

packages/api-browser/window.js
BenLambertNcl: Yeah I think waiting for the rollup PR to go in is probably a good idea, because any other way we do it will just be changed to that as soon as its in anyway.

jleft: Okay, sounds good - if it's easier, I'm happy for this to go in first (without moving `accessibleWindows` out of global scope), then add the necessary changes onto the rollup PR? Up to you, I'm not sure which is less effort :)

BenLambertNcl: Yeah sounds good to me :+1:, Ill leave it global scope for now.

BenLambertNcl:
The `window.opener` reference has no `onclose` event as its not a full window object, so it is not possible to remove the 'parent' from the `accessibleWindows` object (at least, not via that method). There is a `closed` field which returns a boolean if the window is closed which could be used in some way, but it's not as nice.

jleft:
>The `window.opener` reference has no `onclose` event as its not a full window object, so it is not possible to remove the 'parent' from the `accessibleWindows` object (at least, not via that method). There is a `closed` field which returns a boolean if the window is closed which could be used in some way, but it's not as nice.

Maybe the parent could send a message to the child when its closing?

jleft (approved):

jleft:
LGTM, merging now - I'll remove `accessibleWindows` from global scope when I rebase the Rollup/ESE6 modules PR.

PR #38, opened 2017-03-24
Add Rollup and use ES6 Modules

jleft:


packages/api-browser/src/screen-snippet.js
```diff
@@ -1,4 +1,5 @@
 /* globals html2canvas */
+import 'html2canvas';
```
BenLambertNcl, line 2: Could we not do `import html2canvas from 'html2canvas'` here? Then we could remove the eslint comment?

packages/api-electron/package.json
```diff
@@ -4,11 +4,16 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "build": "rimraf dist && rollup -c",
+    "test": "npm run build"
```
BenLambertNcl, line 8: I'm guessing this is used by rollup?

BenLambertNcl (commented):
Looks good. Much nicer than before :+1:

packages/api-electron/package.json
jleft: The test script? It's not used by rollup - Travis CI now runs `lerna run test`, which will run `npm test` in all packages, this is just to make sure the build step isn't broken.

BenLambertNcl: Ah I missed that. :+1:

packages/api-browser/src/screen-snippet.js
jleft: I've tried to, but I don't think so... html2canvas's package.json has a `main` property which references a generated UMD bundle, but it doesn't have a `module` or `jsnext` properties.

There might be a way to tell Rollup to pick up `src/core.js` in html2canvas  (where it is exported), then build it from there, but I can't see how to do that (not even sure if it's possible) - I've had no luck so far... if you can see a way to do it, then please let me know!

BenLambertNcl: I think if it's not simple then we should leave it as is.

BenLambertNcl (approved):

BenLambertNcl (approved):

PR #40, opened 2017-03-27
Window api extension

BenLambertNcl:
Added new methods to the window API:
* Close
* Show (Does not work for browser)
* Hide (Does not work for browser)
* Focus
* Blur

As well as being able to get the current window object by passing no parameters to the constructor (i.e. `new ssf.Window()` returns the current window object) or calling `ssf.Window.getCurrentWindow()`. For the browser, close also only works on windows that were opened via `window.open` (i.e. the window cannot close itself or its parents).

### DIFF
--- a/packages/api-electron/index.js
+++ b/packages/api-electron/index.js
@@ -112,40 +112,31 @@ ipc.on('ssf-get-window-id', (e) => {
 });
 
 ipc.on('ssf-close-window', (e, id) => {
-  const win = BrowserWindow.fromId(id);
-  // Don't need to remove the window from the windows array, the onclose event does that for us
-  if (win) {
-    win.close();
-  }
+  getWindowFromId(id, (win) => win.close());
 });
 
 ipc.on('ssf-show-window', (e, id) => {
-  const win = BrowserWindow.fromId(id);
-  if (win) {
-    win.show();
-  }
+  getWindowFromId(id, (win) => win.show());
 });
 
 ipc.on('ssf-hide-window', (e, id) => {
-  const win = BrowserWindow.fromId(id);
-  if (win) {
-    win.hide();
-  }
+  getWindowFromId(id, (win) => win.hide());
 });
 
 ipc.on('ssf-focus-window', (e, id) => {
-  const win = BrowserWindow.fromId(id);
-  if (win) {
-    win.focus();
-  }
+  getWindowFromId(id, (win) => win.focus());
 });
 
 ipc.on('ssf-blur-window', (e, id) => {
+  getWindowFromId(id, (win) => win.blur());
+});
+
+const getWindowFromId = (id, cb) => {
   const win = BrowserWindow.fromId(id);
   if (win) {
-    win.blur();
+    cb(win);
   }
-});
+};
 
 module.exports.app = {
   ready
--- a/packages/api-electron/index.js
+++ b/packages/api-electron/index.js
@@ -108,7 +108,7 @@ const ready = (cb) => {
 };
 
 ipc.on('ssf-get-window-id', (e) => {
-  e.returnValue = BrowserWindow.fromWebContents(e.sender);
+  e.returnValue = BrowserWindow.fromWebContents(e.sender).id;
 });
 
 ipc.on('ssf-close-window', (e, id) => {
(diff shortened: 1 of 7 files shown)


packages/api-specification/src/window-api.html
```diff
@@ -38,13 +38,30 @@
...
 
   <p>The Window API is used to create a new browser window.</p>
 
   <h3>Constructor</h3>
 
   <pre>
+  new ssf.Window()
+  </pre>
+
+  <p>Creates a new window object for the current window.</p>
```
ColinEberhardt: Just wondering what the use-case for this is?

packages/api-specification/src/window-api-demo.js
```diff
@@ -2,14 +2,45 @@ var newWindowButton = document.getElementById('new-window-test');
...
+    win = new ssf.Window(url, windowName, {
       child: isChild
     });
   });
 };
+
+var closeWindow = document.getElementById('close-window');
+
+closeWindow.onclick = () => {
+  win.close();
```
ColinEberhardt: if appReady never resolved, this will be undefined. I know this shouldn't happen, but perhaps it would be clearer if all of the code within this file sat within the appReady promise `then` block?

packages/api-electron/index.js
```diff
@@ -108,6 +107,46 @@ const ready = (cb) => {
...
+
+ipc.on('ssf-focus-window', (e, id) => {
+  const win = BrowserWindow.fromId(id);
+  if (win) {
+    win.focus();
+  }
+});
+
+ipc.on('ssf-blur-window', (e, id) => {
+  const win = BrowserWindow.fromId(id);
```
ColinEberhardt: Minor refactor, I'd move the following boiler-plate into a helper function:

```
const win = BrowserWindow.fromId(id);
if (win) {
   win.blur();
}
```

ColinEberhardt (changes requested):
A few minor comments, but otherwise looking good 👍

packages/api-specification/src/window-api.html
BenLambertNcl: This is so we could call hide/show etc from the current window. At present, there is no way to get the current window by using the API.

ColinEberhardt: Ah, I can see why that would be useful!

From an API perspective, when you call `new ssf.Window()` you really expect to get a 'new' thing. I think this would make more sense as a 'static' method, how about this - `ssf.Window.getCurrent()`

BenLambertNcl: I agree that `new ssf.Window()` is confusing. There is a static method `getCurrentWindow()` which calls this constructor, as we need the methods etc. Instead, we could have a WindowFactory class which returns new windows, or possibly subclass the window class (into `CurrentWindow` and `ExternalWindow` for example).

packages/api-electron/index.js
BenLambertNcl: I have changed it to something like this:
```
ipc.on('ssf-blur-window', (e, id) => {
  getWindowFromId(id, (win) => win.blur());
});

const getWindowFromId = (id, cb) => {
  const win = BrowserWindow.fromId(id);
  if (win) {
    cb(win);
  }
};
```
Do you think this is better, or did you have something else in mind?

ColinEberhardt: Looks good to me!

packages/api-specification/src/window-api.html
ColinEberhardt: Ah, I hadn't spotted that we already have `getCurrentWindow()`.

In that case, just ignore my comment, as long as `getCurrentWindow` is the documented way to get the 'current' Window, we can just treat the fact that it uses `new ssf.Window()` as an internal implementation detail. Yes, this is also available externally, but no big deal!

BenLambertNcl: Sounds good, I'll update the documentation to reflect that.

ColinEberhardt:
@BenLambertNcl  - let me know when this is ready for reviewing again

BenLambertNcl:
Yeah it's ready now.

ColinEberhardt (approved):

ColinEberhardt:
> Yeah it's ready now.

Great - approved 👍 

Just FYI, in case you didn't know already, GitHub does not notify reviewers (or watchers) when you push new commits to a PR, so when you make changes after a review you have to manually notify pope via  comment.

BenLambertNcl:
Ah I didn't realise that. Will keep in mind for next time. Thanks

PR #41, opened 2017-03-28
Prototype bootstrap logic

ColinEberhardt:
This quick prototype modifies the electron bootstrapping logic to more closely mirror OpenFin.

The `main.js` file has been moved from the demo project into the `api-electron` project, and an executable, `ssf-electron` added that runs this. An `app.json` file, with the same format as OpenFin, is used to define how the application starts up.

This is a bit rough and ready, but looks like a viable way forward to me!

reopens https://github.com/ScottLogic/ssf-desktop-wrapper/issues/32

I've backed out your fix for #32 - on reflection I think there is currently more value in sharing all the same application code between the implementations than being able to run them in parallel.


BenLambertNcl (approved):
Looks really good. As for reopening #32, how about using rollup to insert the correct port? We could use [this plugin](https://github.com/rollup/rollup-plugin-replace).

PR #44, opened 2017-03-30
Window events

BenLambertNcl:
Added window events, which can be subscribed to via `addListener(event, callback)` and removed using `removeListener(event, callback)`. The event strings that are used are the electron events (i.e. `blur`, `hide`, `show`) which get converted to the OpenFin/browser equivalent. Window events on the browser are limited.


ColinEberhardt (approved):
LGTM :+1:

PR #47, opened 2017-03-31
Replace HTML documentation with markdown

BenLambertNcl:
* Added API docs as markdown over html
* Added node script to convert markdown to html and append the html to the demo projects (markdown must have the same filename as demo file. i.e. `notification-api.md` will be appended to `notification-api.html`)
* Reorganized the `api-specification` folder structure (now have directories for app, notifications etc). The links to JavaScript/CSS files do not need to be changed as `copyfiles` flattens the folder structure.


ColinEberhardt (approved):
Very nice :-)

In future, if you find yourself navigating file / directory hierarchies and filtering the results, I'd recommend giving globby a go:

https://github.com/sindresorhus/globby

BenLambertNcl:
That looks really useful. Thanks for the link :)

PR #48, opened 2017-04-03
Replace strings with constants for electron ipc

BenLambertNcl:
Add constants for electron IPC calls. Can be accessed using `constants.ipc.SSF_...`.


AndyMJones (approved):
lgtm :+1:

PR #49, opened 2017-04-04
Add option to make new window frameless

BenLambertNcl:
Only works in Electron and OpenFin. Also added option to make window transparent.

### DIFF
--- a/packages/api-electron/index.js
+++ b/packages/api-electron/index.js
@@ -26,12 +26,13 @@ module.exports = (appJson) => {
   });
 
   ipc.on(constants.ipc.SSF_NEW_WINDOW, (e, msg) => {
-    const options = {
-      webPreferences: {
-        sandbox: true,
-        preload: preloadPath
-      }
+    const webPreferences = {
+      sandbox: true,
+      preload: preloadPath
     };
+
+    const options = msg.features;
+    options.webPreferences = webPreferences;
 
     if (msg.features && msg.features.child) {
       options.parent = BrowserWindow.fromWebContents(e.sender);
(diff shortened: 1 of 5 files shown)


packages/api-electron/index.js
```diff
@@ -25,13 +25,14 @@ module.exports = (appJson) => {
...
-        sandbox: true,
-        preload: preloadPath
-      }
+    const webPreferences = {
+      sandbox: true,
+      preload: preloadPath
     };
 
+    const options = msg.features;
+    options.webPreferences = webPreferences;
```
AndyMJones: I'd probably use `Object.assign` here instead
```
const options = Object.assign(
  {}, 
  message.features, 
  { webPreferences: { sandbox: true, preload: preloadPath } }
);

File: packages/api-openfin/src/window.js [hunk 2]
```diff
@@ -10,23 +10,29 @@ class Window {
...
-      if (features && features.child) {
-        newWindow = new fin.desktop.Window({
+      const windowOptions = features || {};
+      if (windowOptions.transparent) {
+        // OpenFin needs opacity between 1 (not transparent) and 0 (fully transparent)
+        windowOptions.opacity = features.transparent === true ? 0 : 1;
+      }
+
+      if (windowOptions.child) {
+        Object.assign(windowOptions, {
```
AndyMJones: I'm not a fan of overriding `windowOptions`. I'd assign a new `const` and use that instead

BenLambertNcl:
Done :)

AndyMJones (approved):
lgtm

PR #52, opened 2017-04-05
Add postMessage to window API

BenLambertNcl:
Adds `window.postMessage(message)` to the window API. In the browser, this uses `postMessage()`, whereas in OpenFin and Electron, it is just a wrapper around the `MessageService` call.


packages/api-specification/src/messaging/messaging-api.md
```diff
@@ -1,5 +1,5 @@
 ## Overview
-The Messaging API is used to communicate between windows
+The Messaging API is used to communicate between windows. Messages can also be send directly to a window using the window API.
```
AndyMJones: sent*

BenLambertNcl:
Fixed :)

AndyMJones (approved):

PR #53, opened 2017-04-05
Add shared options

AndyMJones:


packages/api-openfin/src/window.js
```diff
@@ -1,43 +1,80 @@
...
-          windowOptions,
+          mergedOptions,
           {
             name,
             url
           }
         );
 
-        newWindow = new fin.desktop.Window(options, () => newWindow.show(), handleError);
+        newWindow = new fin.desktop.Window(childOptions, null, handleError);
```
BenLambertNcl: We could just have `newWindow = new fin.desktop.Window(childOptions);`. We need to add window success/error callbacks at some point anyway

packages/api-specification/src/window/window-api.md
```diff
@@ -26,15 +26,23 @@ Object to set features of the new window (non browser only).
 * `alwaysOnTop: boolean` - whether the window is always on top of other windows.
 * `backgroundColor: string` - set the backgroundColor of the window before the page loads.
 * `child: boolean` - make the new window a child of the current window.
+* `center: boolean` - position the new window in the center of the main screen. (only works on new windows with new ids)
 * `frame: boolean` - create the new window with a frame.
+* `hasShadow: boolean` - show shadows for frameless windows.
+* `height: Integer` - window's height in pixels.
```
BenLambertNcl: Small 'i'? Or all the others capitals? Don't mind which but think it will look better if its consistent :)

BenLambertNcl (commented):
Looks good, some very minor comments :+1:

packages/api-openfin/src/window.js
```diff
@@ -1,43 +1,80 @@
 let currentWindow = null;
 import MessageService from './message-service';
 
+const optionsMap = {
```
BenLambertNcl: If this is at the top its probably best to have the eventMap at the top as well

BenLambertNcl (approved):

PR #54, opened 2017-04-06
Change window methods to be promise based

BenLambertNcl:
Change methods in the window API to return promises. This normalizes the APIs to all be asynchronous.

### DIFF
--- a/packages/api-browser/src/window.js
+++ b/packages/api-browser/src/window.js
@@ -5,7 +5,18 @@ import {
 
 let currentWindow = null;
 
-const windowClosedError = 'Error: The window does not exist or the window has been closed';
+const windowClosedError = new Error('The window does not exist or the window has been closed');
+
+const withInnerWindow = (fn) => {
+  return new Promise((resolve, reject) => {
+    if (this.innerWindow) {
+      fn(this.innerWindow);
+      resolve();
+    } else {
+      reject(windowClosedError);
+    }
+  });
+};
 
 class Window {
   constructor(...args) {
@@ -28,14 +39,7 @@ class Window {
 
   close() {
     // Close only works on windows that were opened by the current window
-    return new Promise((resolve, reject) => {
-      if (this.innerWindow) {
-        this.innerWindow.close();
-        resolve();
-      } else {
-        reject(windowClosedError);
-      }
-    });
+    return withInnerWindow(innerWindow => innerWindow.close());
   }
 
   show() {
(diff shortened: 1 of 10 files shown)


packages/api-browser/src/window.js
```diff
@@ -26,9 +28,14 @@ class Window {
 
   close() {
     // Close only works on windows that were opened by the current window
-    if (this.innerWindow) {
-      this.innerWindow.close();
-    }
+    return new Promise((resolve, reject) => {
```
ColinEberhardt: How about refactoring to remove some of the repetition? For example:

```
const withInnerWindow(fn) =>
  new Promise((resolve, reject) => {
      if (this.innerWindow) {
        fn(this.innerWindow);
        resolve();
      } else {
        reject(windowClosedError);
      }
    });
```

Then `close` becomes the following:

```
  close() {
    return withInnerWindow(innerWindow => innerWindow.close());
  }
```

packages/api-electron/index.js
```diff
@@ -112,40 +127,85 @@ const ready = (cb) => {
...
+ipc.on(IPC_SSF_GET_WINDOW_ID, (e) => {
   e.returnValue = BrowserWindow.fromWebContents(e.sender).id;
 });
 
-ipc.on(constants.ipc.SSF_CLOSE_WINDOW, (e, id) => {
-  getWindowFromId(id, (win) => win.close());
+ipc.on(IPC_SSF_CLOSE_WINDOW, (e, id, nonce) => {
+  const messageInfo = [e.sender, IPC_SSF_CLOSE_WINDOW, nonce];
+
+  getWindowFromId(id, (win) => {
```
ColinEberhardt: You've got a similarly repeated pattern here that could be factored out.

packages/api-specification/src/window/window-api.md
```diff
@@ -54,15 +54,15 @@ A reference to the new window.
 
 ### Class Methods
 * `addListener(event, callback)` - Add an event listener to execute a given callback
-* `blur()` - Remove focus from the window
-* `close()` - Close the window
-* `focus()` - Give the window focus
-* `hide()` - Hide a visible window (Non browser only)
+* `blur()` - Remove focus from the window. Returns a promise
```
ColinEberhardt: You should probably mention what the promise resolves to? Even if it resolves to nothing!

ColinEberhardt (changes requested):

BenLambertNcl:
Ready for re-review. I have added a small change to the OpenFin API to run the reject function when the window has been closed, keeping it in line with the Electron API.

ColinEberhardt (approved):

PR #59, opened 2017-04-12
Additional window methods

AndyMJones:
These are just the "easier" methods to implement between Openfin and Electron there are some others that  could still possibly be implemented. 

Only the Electron methods have been implemented across the two however in the future we may want to also have the Openfin methods in both.


packages/api-openfin/src/window.js
```diff
@@ -136,27 +138,227 @@ class Window {
...
+  }
+
+  getMinimumSize() {
+    return this.getOptions()
+      .then((options) => {
+        return [options.minWidth, options.minHeight];
+      });
+  }
+
+  getOptions() {
```
AndyMJones, line 193: This isnt implemented in Electron  but is used a lot within the Openfin methods. Is it acceptable to have this method usable by the user?

packages/api-openfin/src/window.js
```diff
@@ -136,27 +138,227 @@ class Window {
...
+  }
+
+  getTitle() {
+    return this.getOptions()
+      .then((options) => {
+        return options.title;
+      });
+  }
+
+  getState() {
```
AndyMJones, line 230: Ditto

packages/api-electron/index.js
```diff
@@ -10,16 +10,48 @@ const {
...
+  IPC_SSF_SET_WINDOW_RESIZABLE,
+  IPC_SSF_SET_WINDOW_SIZE,
+  IPC_SSF_SET_WINDOW_SKIP_TASKBAR,
   IPC_SSF_SEND_MESSAGE,
   IPC_SSF_SHOW_WINDOW,
   IPC_SSF_WINDOW_EVENT,
-  IPC_SSF_WINDOW_SUBSCRIBE_EVENTS
+  IPC_SSF_WINDOW_HAS_SHADOW,
+  IPC_SSF_WINDOW_SUBSCRIBE_EVENTS,
+  IPC_SSF_UNMAXIMIZE_WINDOW
```
BenLambertNcl: What do you think about having a general `const { constants } = require('constants')` as well as the individual ones so we can import all of these in one go (as there's so many)?

packages/api-electron/src/preload/window.js
```diff
@@ -4,13 +4,45 @@ import {
...
+  IPC_SSF_SET_WINDOW_POSITION,
+  IPC_SSF_SET_WINDOW_RESIZABLE,
+  IPC_SSF_SET_WINDOW_SIZE,
+  IPC_SSF_SET_WINDOW_SKIP_TASKBAR,
   IPC_SSF_SHOW_WINDOW,
   IPC_SSF_WINDOW_EVENT,
-  IPC_SSF_WINDOW_SUBSCRIBE_EVENTS
+  IPC_SSF_WINDOW_HAS_SHADOW,
+  IPC_SSF_WINDOW_SUBSCRIBE_EVENTS,
+  IPC_SSF_UNMAXIMIZE_WINDOW
```
BenLambertNcl: Ditto

packages/api-electron/index.js
```diff
@@ -127,47 +159,220 @@ const ready = (cb) => {
...
+  const messageInfo = [e.sender, IPC_SSF_MAXIMIZE_WINDOW, nonce];
+  callWindowFunction(id, messageInfo, win => win.maximize());
+});
+
+ipc.on(IPC_SSF_MINIMIZE_WINDOW, (e, id, nonce) => {
+  const messageInfo = [e.sender, IPC_SSF_MINIMIZE_WINDOW, nonce];
+  callWindowFunction(id, messageInfo, win => win.minimize());
+});
+
+ipc.on(IPC_SSF_UNMAXIMIZE_WINDOW, (e, id, nonce) => {
```
BenLambertNcl: We should probably split up these ipc events into another file and require them in here. Probably for another PR though

packages/api-openfin/src/window.js
```diff
@@ -52,19 +52,25 @@ const optionsMap = {
...
+  RESTORED: 'restored'
+};
+
 const checkWindowOpen = (win, windowMethod, reject) => {
   if (win) {
-    windowMethod();
+    return windowMethod();
   } else {
-    reject();
+    return reject(new Error('window does not exist'));
```
BenLambertNcl: The error in electron is `The window does not exist or the window has been closed`. Think they should both be the same, either way is fine

packages/api-openfin/src/window.js
BenLambertNcl: I think its fine, as long as we document the correct way to do it

packages/api-openfin/src/window.js
BenLambertNcl: ditto

packages/api-openfin/src/window.js
```diff
@@ -136,27 +138,227 @@ class Window {
...
+  }
+
+  getState() {
+    return new Promise((resolve, reject) => checkWindowOpen(this.innerWindow, () => this.innerWindow.getState(resolve, reject), reject));
+  }
+
+  hasShadow() {
+    return this.getOptions()
+      .then((options) => {
+        console.log(options);
```
BenLambertNcl: Remove

packages/api-specification/src/window/window-api.md
```diff
@@ -56,13 +56,45 @@ A reference to the new window.
 * `addListener(event, callback)` - Add an event listener to execute a given callback
 * `blur()` - Remove focus from the window. Returns a promise that resolves to nothing
 * `close()` - Close the window. Returns a promise that resolves to nothing
+* `flashFrame(flag)` - Flashes the window’s frame and taskbar icon.
```
BenLambertNcl: Think we need to put that they all/most return a promise and what they resolve to. If they all resolve to nothing, it might be worth saying something like this above the list: `All methods return a promise that resolves to nothing, unless stated otherwise`

packages/api-specification/src/window/window-api.md
```diff
@@ -56,13 +56,45 @@ A reference to the new window.
...
+* `getMinimumSize()` - Returns `Integer[width, height]` - Contains the window’s minimum width and height.
+* `getParentWindow()` - Returns the parent window.
+* `getPosition()` - Returns `Integer[x, y]` - Contains the window’s current position.
+* `getSize()` - Returns `Integer[width, height]` - Contains the window’s width and height.
+* `getTitle()` - Returns the title of the native window.
+* `hasShadow()` - Returns whether the window has a shadow.
 * `hide()` - Hide a visible window (Non browser only). Returns a promise that resolves to nothing
+* `isAlwaysOnTop()` - Returns whether the window is always on top of other windows.
+* `isMaximizable()` - Returns whether the window can be manually maximized by user.
+* `isMaximized()` - Returns whether the window is maximized.
```
BenLambertNcl: Probably want the return type. i.e. `Returns a boolean to state whether ...`

BenLambertNcl (commented):
Good work! Just some small comments :)

packages/api-electron/index.js
AndyMJones: :+1:

BenLambertNcl (approved):

PR #61, opened 2017-04-12
Add compatability matrix

AndyMJones:
This md file shows what methods and options we have implemented and how they map to Electron and OpenFin. The remaining Electron and OpenFin methods and options have also been added.


BenLambertNcl (approved):
Add a link to this in the wiki :+1:

PR #62, opened 2017-04-13
Add window success and error callbacks

AndyMJones:


packages/api-browser/src/window.js
```diff
@@ -19,34 +19,37 @@ const withInnerWindow = (win, fn) => {
...
 
 class Window {
-  constructor(...args) {
-    if (args.length === 0) {
+  constructor(options, callback, errorCallback) {
+    this.children = [];
+
+    this.eventListeners = new Map();
+
+    if (options === undefined || options === null) {
```
ColinEberhardt: I'd just check for falsy, i.e. `!options`

packages/api-electron/src/preload/window.js
```diff
@@ -12,29 +12,38 @@ const generateNonce = () => {
...
+
+    this.eventListeners = new Map();
+    MessageService.subscribe('*', 'ssf-window-message', (...args) => {
+      const event = 'message';
+      if (this.eventListeners.has(event)) {
+        this.eventListeners.get(event).forEach(listener => listener(...args));
+      }
+    });
+
+    if (options === undefined || options === null) {
```
ColinEberhardt: ditto

packages/api-openfin/src/window.js
```diff
@@ -136,6 +95,42 @@ class Window {
         this.eventListeners.get(event).forEach(listener => listener(...args));
       }
     });
+
+    if (options === undefined || options === null) {
```
ColinEberhardt: ditto

packages/api-specification/src/window/window-api.md
```diff
@@ -4,7 +4,7 @@ The Window API is used to create a new browser window. Compatability between sup
 
 ### Constructor
 ```javascript
-new ssf.Window(url, name, features)
+new ssf.Window(features[, callback, errorCallback])
```
ColinEberhardt: Can you please document the new callback arguments.

At this point it is probably worth thinking about when `callback` is actually invoked? can I send messages to the new window within this callback? Or do I have to wait for the `ready` lifecycle event to be called in the new window?

I don't think it's worth trying to normalise this behaviour just yet, but it is probably worth just discovering what the differences are and noting them.

ColinEberhardt (approved):
Some minor stylistic stuff, but otherwise LGTM.

PR #64, opened 2017-04-19
globstars need to be quoted to work on a mac

ColinEberhardt:
see: https://www.npmjs.com/package/copyfiles#command-line

Yes Mr. Autocorrect, that's 'globstar', not 'lobster' :crab:


AndyMJones (approved):
lgtm

PR #67, opened 2017-04-24
Some refactoring

ColinEberhardt:
There was quite a bit of boiler-plate within the OpenFin window implementation in order to adapt the API to Promise-based. Within https://github.com/ColinEberhardt/ssf-desktop-wrapper/commit/01c50fb8b2cd8db9ede941d51f957592bcd5f7dd I've created a generic `asPromise` function that adapts the OpenFin API methods.

Other very minor things, make use of implicit returns with arrow functions, e.g. `f => { return 'bar' }` can be replaced with `f => 'bar'`.

I've also changed the way options are mapped so that un-used options are not passed through to OpenFin. Another minor tweak, I've moved the constant that defines the mappings into the function that uses it (I prefer to restrict the scope of constants / variables as much as possible - I'll let the JIT compiler worry about optimising this!)

### DIFF
--- a/packages/api-openfin/src/window.js
+++ b/packages/api-openfin/src/window.js
@@ -62,9 +62,9 @@ const convertOptions = (options) => {
   };
 
   Object.keys(optionsMap).forEach((optionKey) => {
-    const openfinOptionKey = optionsMap[optionKey];
+    const openFinOptionKey = optionsMap[optionKey];
     if (clonedOptions[optionKey]) {
-      clonedOptions[openfinOptionKey] = clonedOptions[optionKey];
+      clonedOptions[openFinOptionKey] = clonedOptions[optionKey];
       delete clonedOptions[optionKey];
     }
   });
@@ -164,7 +164,7 @@ class Window {
   }
 
   getBounds() {
-    return this.getBounds('focus')
+    return this.asPromise('getBounds')
       .then(bounds => ({
         x: bounds.left,
         y: bounds.top,
--- a/packages/api-openfin/src/window.js
+++ b/packages/api-openfin/src/window.js
@@ -29,29 +29,6 @@ const eventMap = {
   'show': 'shown'
 };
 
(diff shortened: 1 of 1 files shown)


packages/api-openfin/src/window.js
```diff
@@ -135,38 +129,48 @@ class Window {
...
 
   focus() {
-    return new Promise((resolve, reject) => checkWindowOpen(this.innerWindow, () => this.innerWindow.focus(resolve, reject), reject));
+    return this.asPromise('focus');
   }
 
   getBounds() {
-    return new Promise((resolve, reject) => checkWindowOpen(this.innerWindow, () => this.innerWindow.getBounds((bounds) => {
-      resolve({
+    return this.getBounds('focus')
```
BenLambertNcl: Should this not be `this.asPromise('getBounds')`?

packages/api-openfin/src/window.js
```diff
@@ -106,23 +105,18 @@ class Window {
 
     MessageService.subscribe('*', 'test2', (message) => console.log(message));
 
-    const convertedOptions = convertOptions(options);
-    const mergedOptions = Object.assign(
-      {},
-      convertedOptions,
-      options
-    );
+    const openFinOptions = convertOptions(options);
```
BenLambertNcl, line 108: [this constant](https://github.com/ScottLogic/containerjs/pull/67/files#diff-e05e841065008af4a1271781d72f7dbbR65) and this constant should have consistent naming (i.e. `openfin` vs `openFin`)

BenLambertNcl (commented):
Some small comments, but looking good :+1:

ColinEberhardt:
Thanks @BenLambertNcl - I've addressed those comments (and made sure the build passes!)

BenLambertNcl (commented):
LGTM :+1:

BenLambertNcl (approved):

PR #60, opened 2017-04-12
Add Spectron test framework for OpenFin and Electron

BenLambertNcl:
Adds Spectron testing framework for OpenFin and Electron. Tests do not currently run on CI, so a separate PR once that is figured out. Browser tests will also be added in another PR.

PR #70, opened 2017-04-26
Add spectron test framework

BenLambertNcl:
This PR adds the spectron test suite to run tests against Electron and OpenFin.
* AppVeyor is needed to run the tests on CI (at least for now, as OpenFin tests do not work on linux/mac)
* Lerna was updated to RC3, as it had better reliability on CI than the beta version we were using (lerna bootstrap kept failing)

Currently, to run the tests, use `npm run test:ui` in the top level directory.

This also closes #60.


packages/api-openfin-demo/package.json
```diff
@@ -8,19 +8,26 @@
...
     "start": "npm-run-all --parallel serve launch",
     "postinstall": "rimraf src && copyfiles -f \"node_modules/containerjs-api-specification/src/**\" src && copyfiles -f \"node_modules/containerjs-api-openfin/dist/**\" src",
-    "test": "echo \"No test specified\" && exit 0"
+    "test": "exit 0",
+    "test:ui": "mocha test/main.spec.js"
   },
   "author": "",
   "license": "Apache-2.0",
   "dependencies": {
+    "assert": "^1.4.1",
```
ColinEberhardt: there are lot of additional dependencies being added to the various projects. Could you just add the test-specific dependencies to `containerjs-api-tests`? these dependencies should propagate through to the projects that depend on `containerjs-api-tests`

packages/api-openfin/src/window.js
```diff
@@ -65,7 +65,9 @@ const convertOptions = (options) => {
     const openFinOptionKey = optionsMap[optionKey];
     if (clonedOptions[optionKey]) {
       clonedOptions[openFinOptionKey] = clonedOptions[optionKey];
-      delete clonedOptions[optionKey];
+      if (openFinOptionKey !== optionKey) {
```
ColinEberhardt: well spotted!

packages/api-openfin/src/window.js
```diff
@@ -103,8 +105,6 @@ class Window {
       return this;
     }
 
-    MessageService.subscribe('*', 'test2', (message) => console.log(message));
```
ColinEberhardt: I did wonder whether that code was really supposed to be there when I refactored!

packages/api-tests/test/window.spec.js
```diff
@@ -0,0 +1,92 @@
...
+        });
+      `;
+      return executeAsyncJavascript(app.client, script).then((result) => {
+        return app.client.getWindowCount().then((count) => {
+          assert.equal(count, 2);
+        });
+      });
+    });
+
+    it('Check new window has correct x position', function() {
```
ColinEberhardt, line 74: This seems like a lot of code for a simple verification - do you think we'll see a similar quantity of code for other tests?

ColinEberhardt (changes requested):
Well done on getting this up and running!

A few quick things ..

Can you separate out your bug fixes into a separate PR? They are simple fixes so should get merged straight away!

My main concern with these tests is the complexity of the test cases themselves. I think it is very important to ensure that the overhead of adding / maintaining tests is not that big. I'd hope that the test code is not that much more complicated than the code it is testing!

packages/api-tests/test/window.spec.js
BenLambertNcl: I think for tests involving multiple windows, yes. This is because we probably want to run JavaScript in window 1, then switch to window 2, then run JavaScript in window 2, then check for some value/action. Maybe we could extract most of the JavaScript string that gets passed in some way? Something like:
```
const windowMethod = (method) => {
   exectueJavaScript(client, '
     var callback = ...
     ssf.Window.getCurrent().${method}().then(() => callback())
  ')
}
```
Calling it with `windowMethod('getBounds')`, which would at least reduce a lot of the boiler plate in the tests somewhat?

ColinEberhardt: It's worth a try. I think it's worth working on this test case to see what you could move into code that could be common across multiple tests. Why not add a case that verifies the y position as well?

(although in practice I'd verify both in one test, but at least it helps highlight the problem!)

BenLambertNcl:
PR has been updated. Really like the way the tests look now, but they still seem to be doing a lot (but not sure this can be help as explained before)

ColinEberhardt (approved):
That is *much* better - nice job :-)

BenLambertNcl:
I don't have permission to add AppVeyor to the repo, an admin will need to add it at some point :)

BenLambertNcl:
Just realised the build failed on AppVeyor as mocha was missing (oops), the final commit adds it back in

PR #71, opened 2017-04-27
Bugfixes for ipc calls and OpenFin window API

BenLambertNcl:
* A bug was found in the OpenFin API where it was deleting any options that were the same in the map (e.g. child) which was causing the tests to fail. This has been fixed.
* Fixed ipc.once listeners taking 1 parameter, not 2


ColinEberhardt (approved):
👍

PR #75, opened 2017-04-28
Change to run all tests in api-tests

BenLambertNcl:
The Spectron tests were running inside the demo projects. This PR moves the tests to be ran inside the api-tests directory. An environment variable is used to differentiate between containers (MOCHA_CONTAINER)


packages/api-tests/README.md
```diff
@@ -0,0 +1,8 @@
+# ContainerJS Testing
+This project contains the Spectron UI tests for the OpenFin and Electron APIs
+
+## Testing
+```
+npm i
+npm run test:ui
```
ColinEberhardt, line 7: Rather than `test:ui` why not just use `test`, so that this become `npm test`?

ColinEberhardt (approved):
One minor comment - but otherwise looks good to me 👍

packages/api-tests/README.md
BenLambertNcl: This was because if it was npm test, lerna would run this when we run on travis (which would fail). We can change this if we manage to get OpenFin tests to work on linux.

ColinEberhardt: Ah - fair point - might be worth adding a quick note to this readme with that information.

BenLambertNcl:
Updated the README :)

PR #77, opened 2017-05-02
Add tests for message service

BenLambertNcl:
Added tests for the messaging service, mainly testing the logic for sending and subscribing to particular windows and topics. Tests were added for the message service to make sure that multiple parts of the API could be tested with Spectron.

### DIFF
--- a/packages/api-tests/test/messageing.spec.js
+++ b/packages/api-tests/test/messageing.spec.js
@@ -1,8 +1,5 @@
 const assert = require('assert');
-const liveServer = require('live-server');
 const {
-  before,
-  after,
   beforeEach,
   afterEach,
   describe,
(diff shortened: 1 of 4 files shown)


ColinEberhardt (approved):
Looks good to me. A couple of thoughts ...

Is there some common config / setup code from the Windows tests that could be put in a common location?

You could probably remove the repetitive promise chaining with something like the following:

```
const chainPromises = (promises) => promises.reduce((acc, cur) => acc.then(curr);
}, promises[0]);
```

Which you'd use as follows:
```
const steps = [
  () => selectWindow(1),
  () => executeStuff(),
  ()  => selectWindow(2)
]
return chainPromises(steps);
```

(I've not tested that!)

BenLambertNcl:
Regarding the first point, yes there are some thing that are repeated and can be moved (for example, `executeAsyncJavascript`). Would `test-helpers.js` be a suitable name (I know a lot of people dislike the name 'helper' or similar).

I will take a look at your second suggestion. For me though, I think the promise chaining reads slightly better (but that's personal preference I suppose :) )

AndyMJones (approved):
Looks good to me but I agree with Colin that that the steps way looks a little bit nicer if it works.

BenLambertNcl:
Ill look at adding it in then :)

ColinEberhardt:
> Ill look at adding it in then :)

Surely you mean you'll look at adding it in then().then().then().then()?

ColinEberhardt:
> Ill look at adding it in then :)

Surely you mean you'll look at adding it in then().then().then().then()?

BenLambertNcl:
Made those changes.

PR #78, opened 2017-05-03
Add gh-pages

AndyMJones:


docs/Gemfile.lock
```diff
@@ -0,0 +1,201 @@
+GEM
```
BenLambertNcl: Don't think we should need to track this file?

BenLambertNcl (approved):
LGTM :+1:, Just one file that doesn't need tracking I think

ColinEberhardt:
LGTM - although does this content now duplicate some of the documentation in the `api-specification` package?

BenLambertNcl:
I have a PR ready to go in once this is merged to copy the docs folder into each of the API demos, and remove the docs from `api-specification`.

PR #80, opened 2017-05-03
Remove duplication of API documentation

BenLambertNcl:
Removes old API documentation and copies from the docs folder used on Github pages


copy-markdown.js
```diff
@@ -11,7 +11,9 @@ directories.forEach((dir) => {
     if (path.extname(file) === '.md') {
       // Convert markdown to html
       const md = fs.readFileSync(path.join(srcDir, file), {encoding: 'utf8'});
-      const html = marked.parse(md);
+      // Need to strip out the file data used for gh-pages
+      const strippedMd = md.split('---')[2];
```
AndyMJones: This seems risky because if we have a table further down the document, we are only going to get the file between the start and the table. My first though was to just concat everything from 2 onward but split also removes '---' so that wont work either. I'd look into either removing it once its html or use indexOf to find the second occurance and then slice after that.

AndyMJones (changes requested):

copy-markdown.js
BenLambertNcl: Good spot! :+1:

AndyMJones (approved):
lgtm

PR #82, opened 2017-05-04
Add electron remote module to preload scripts

BenLambertNcl:
Adds the electron remote module to the preload scripts. This allows a much nicer way of accessing the main process objects (especially for the window API). This removes a lot of the IPC code we had and makes things like the event listeners much nicer. We can also now get the full `BrowserWindow` object in the renderer process, not just a proxy/id. The reason this was not implemented initially was due to the remote module not being enabled in the sandbox renderer, which now isn't the case.

Some IPC calls remain, such as creating a new window so we keep a reference to all windows so they are not garbage collected and the message service.


AndyMJones (approved):
Having a try catch inside a promise still feels weird to me, especially if you are just rejecting the error. I guess the user can just add a catch to each one of their method calls rather than us dealing with it, but that still makes me think we don't need the try catch.

ColinEberhardt:
Nice ...
screen shot 2017-05-05 at 08 49 53

BenLambertNcl:
@JohnEz yes you are right, it works without the try catch. Ill update the PR

PR #84, opened 2017-05-05
Add tests for notification API

BenLambertNcl:


BenLambertNcl:
During discussions offline, it was decided notification tests were too tightly coupled to the implementation of the notifications. Closing this due to it no longer being needed.

PR #85, opened 2017-05-05
Add window method tests

AndyMJones:


packages/api-tests/test/window.spec.js
```diff
@@ -54,16 +54,328 @@ describe('Window API', function(done) {
...
+        assert.equal(result.value.width, width)
+        assert.equal(result.value.x, x)
+        assert.equal(result.value.y, y)
+      }
+    ];
+
+    return chainPromises(steps);
+  });
+
+  it('Should get if the window is alwaysOnTop', function() {
```
BenLambertNcl: Guessing there's a word missing here?

packages/api-tests/test/window.spec.js
```diff
@@ -54,16 +54,328 @@ describe('Window API', function(done) {
...
+      () => selectWindow(app.client, 1),
+      () => app.client.waitForVisible('.visible-check'),
+      () => callWindowMethod('isAlwaysOnTop'),
+      (result) => assert.equal(result.value, alwaysOnTop)
+    ];
+
+    return chainPromises(steps);
+  });
+
+  it('Should get if the window is maximizable', function() {
```
BenLambertNcl: ditto. There are a few more too further down

packages/api-tests/test/window.spec.js
```diff
@@ -54,16 +54,328 @@ describe('Window API', function(done) {
     return executeAsyncJavascript(app.client, script);
   });
 
+  it('Should close the window', function() {
```
BenLambertNcl: Think having all of these in a describe would be nice so we can group the window option tests together, the window method tests together etc

packages/api-tests/test/window.spec.js
```diff
@@ -54,16 +54,328 @@ describe('Window API', function(done) {
...
+    const windowTitle = 'windownameclose';
+    const windowOptions = {
```
BenLambertNcl: Could have a global `defaultWindowOptions` and just override what we need for each test?

BenLambertNcl (commented):
Few comments, but looking good :+1:

BenLambertNcl (approved):

ColinEberhardt (commented):
There's quite a bit of repeated code in the tests. Could you create a utility function that constructs a window ready for testing?

```
const createWindow(windowTitle) => {
      const windowOptions = getWindowOptions({
        name: windowTitle
      });

      const steps = [
        () => app.client.isVisible('.visible-check'),
        () => openNewWindow(app.client, windowOptions),
        () => selectWindow(app.client, 1),
        () => app.client.waitForVisible('.visible-check')
    ]
    return chainPromises(steps);
}
```

PR #94, opened 2017-05-08
Fix for OpenFin displaying undefined as string

BenLambertNcl:
`document.getElementById('message').innerText = undefined` displays the string undefined, not an empty string.


AndyMJones (approved):
lgtm

PR #105, opened 2017-05-10
Add additional window tests

AndyMJones:
Some of the tests are currently set to skip to avoid broken tests (code needs fixing rather than test) or not yet implemented tests

### DIFF
--- a/packages/api-openfin/src/window.js
+++ b/packages/api-openfin/src/window.js
@@ -35,6 +35,12 @@ const windowStates = {
   RESTORED: 'restored'
 };
 
+const DEFAULT_X = 560;
+const DEFAULT_Y = 220;
+
+const DEFAULT_WIDTH = 800;
+const DEFAULT_HEIGHT = 600;
+
 const convertOptions = (options) => {
   let clonedOptions = Object.assign({}, options);
 
@@ -63,7 +69,7 @@ const convertOptions = (options) => {
 
   Object.keys(optionsMap).forEach((optionKey) => {
     const openFinOptionKey = optionsMap[optionKey];
-    if (clonedOptions[optionKey]) {
+    if (clonedOptions[optionKey] != null) {
       clonedOptions[openFinOptionKey] = clonedOptions[optionKey];
       if (openFinOptionKey !== optionKey) {
         delete clonedOptions[optionKey];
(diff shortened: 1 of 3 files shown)


packages/api-openfin/src/window.js
```diff
@@ -196,18 +216,25 @@ class Window {
     return this.asPromise('getOptions');
   }
 
+  getFrameSize() {
```
BenLambertNcl: Were we not removing this with `getOptions` still returning the original value?

packages/api-tests/test/window.spec.js
```diff
@@ -106,6 +137,254 @@ describe('Window API', function(done) {
...
+      const steps = [
+        ...setupWindowSteps(windowOptions),
+        () => callAsyncWindowMethod('getMinimumSize'),
+        (result) => assert.equal(result.value[1], minHeight + frameSize)
+      ];
+
+      return chainPromises(steps);
+    });
+
+    it.skip('Should return the parent window', function() {
```
BenLambertNcl, line 252: Should this be skipped? `getParentWindow` was fixed right?

packages/api-tests/test/window.spec.js
```diff
@@ -176,39 +455,41 @@ describe('Window API', function(done) {
...
+        () => callAsyncWindowMethod('isResizable'),
         (result) => assert.equal(result.value, true)
       ];
 
       return chainPromises(steps);
     });
 
-    it('Should return a boolean stating if the window is maximized', function() {
-      const windowTitle = 'windownamemaximized';
+    it.skip('Should load a url', function() {
```
BenLambertNcl, line 480: Why is this skipped?

packages/api-tests/test/window.spec.js
AndyMJones: From my testing it all works, however the object it returns is too large which causes the test to fail.

packages/api-tests/test/window.spec.js
AndyMJones: I can't work out why this test doesn't work if I'm honest. I've tried the method in the demo and it works, however in this test the page never loads and the test gets stuck waiting for the promise to resolve. It feels like it might be a problem with our testing setup so i have just left it as skip until we know if it is testable or not.

BenLambertNcl (approved):

ColinEberhardt:
Looks good :-)

For the skipped tests, it's probably worth capturing why they are skipped as an issue.

PR #109, opened 2017-05-10
Only limit lerna bootstrap concurrency on CI

jleft:
This change allows Lerna to parallelise the bootstrap process when not running on CI - so it should be around 2-4x faster locally. Not limiting the concurrency on CI caused sporadic failures, so remains unchanged.


ColinEberhardt (approved):
👍

BenLambertNcl (approved):

PR #113, opened 2017-05-11
Add test coverage matrix

### LINKED ISSUE #91: Matrix of test coverage
It would be great to have a matrix which indicates the overall test coverage of the APIs. Similar to this sort of thing:

https://kangax.github.io/compat-table/es6/

With this, plus our compatability tables, and API documentation, it does feel like it should be possible to code-gen the various outputs.

BenLambertNcl:
This adds a test matrix to the documentation. Closes #91.

_Note:_ The process is currently manual. Any changes to the tests require you to run `npm run test:ci` and commit the new test matrix. This will be improved once we add the auto generating documentation, so everything can be generated on CI.

What the matrix will look like:
image


ColinEberhardt:
That looks great :tada:

packages/api-tests/generate-test-martix.js
```diff
@@ -0,0 +1,105 @@
+const fs = require('fs');
```
ColinEberhardt: filename martix => matrix

ColinEberhardt (commented):
This looks really good. The only thing I'd ask is that as our build process is becoming a little more complex, I do think it would benefit from some (very lightweight) documentation. Just a few bullet points about what the steps are, what each step does, and any outputs they create

BenLambertNcl:
Yes I agree. I will add the documentation update as a new pull request.

.gitignore
```diff
@@ -67,6 +67,10 @@ npm
 /packages/api-specification/src/bootstrap.min.css
 dist/
 
+# Test output
+/packages/api-tests/electron.json
```
jleft: Could create a directory for this output, to simplify this and give some more context (i.e. if I see an file called `electron.json`, I'm not sure what it contains, but if it's in a directory, like `coverage` or `config`, then I'd have a better idea what it is).

packages/api-tests/package.json
```diff
@@ -5,8 +5,9 @@
   "main": "index.js",
   "scripts": {
     "test:ui": "npm run test:electron && npm run test:openfin",
+    "test:ci": "npm run test:electron -- --reporter reporter.js && npm run test:openfin -- --reporter reporter.js",
     "test:electron": "rimraf src && copyfiles -f \"demo/**\" src && cross-env MOCHA_CONTAINER=electron mocha test/*.js",
-    "test:openfin": "rimraf src && copyfiles -f \"demo/**\" src && copyfiles -f \"node_modules/containerjs-api-openfin/dist/**\" src && cross-env MOCHA_CONTAINER=openfin mocha test/*.js"
+    "test:openfin": "rimraf src && copyfiles -f \"demo/**\" src && copyfiles -f \"node_modules/containerjs-api-openfin/dist/**\" src && cross-env MOCHA_CONTAINER=openfin mocha test/*.js --reporter reporter.js"
```
jleft: Should this be changed?

packages/api-tests/reporter.js
```diff
@@ -0,0 +1,62 @@
+var mocha = require('mocha');
+module.exports = MyReporter;
+const testContainer = process.env.MOCHA_CONTAINER;
+var fs = require('fs');
+var path = require('path');
```
jleft: Nitpick: could group all `require` statements at the top of the file

packages/api-tests/reporter.js
```diff
@@ -0,0 +1,62 @@
...
+
+let jsonOutput = [];
+const pattern = /(#[\w.()]+[\s]?)+/;
```
jleft: Maybe give the variable a more descriptive name (indicating what it is a pattern for)? 

Is the whitespace expression (`[\s]?`) needed?

packages/api-tests/reporter.js
```diff
@@ -0,0 +1,62 @@
...
+
+function MyReporter(runner) {
```
jleft: Nitpick: maybe give a more descriptive name for what the reporter does and move the reporter into a directory, so the root of the package doesn't become cluttered.

packages/api-tests/reporter.js
```diff
@@ -0,0 +1,62 @@
...
+      container: testContainer,
+      tags: tags === null ? [] : tags[0].split(' ')
+    });
+    console.log('%s %s -- error: %s', mocha.reporters.Base.symbols.err, test.title, err.message);
+  });
+
+  runner.on('end', function() {
+    console.log('pass:%d pending:%d failure:%d', passes, pending, failures);
+    fs.writeFileSync(path.join(process.cwd(), testContainer + '.json'), JSON.stringify(jsonOutput));
+    process.exit(failures);
```
jleft: Is this line needed? I think it would be unusual to vary the exit code based on the number of test failures?

jleft (commented):
Looks really good 👍  just added a few comments/questions

packages/api-tests/reporter.js
BenLambertNcl: The whitespace is needed in case we have multiple tags. If we only want max 1 tag per method then I can remove it?

packages/api-tests/reporter.js
BenLambertNcl: This is so `npm test` fails when there is a test failure. Could do something like `process.exit( failures > 0 ? -1 : 0 )`?

packages/api-tests/reporter.js
BenLambertNcl: Actually, I think multiple tags will be necessary when we want specific container tests (i.e. `#electron-only`) so I think we should keep the space

ColinEberhardt (approved):

PR #114, opened 2017-05-12
Add typescript to Electron and OpenFin

AndyMJones:
This PR is just to add typescript to the packages, a future PR should be to replace a lot of the `any`s used with actual types.


packages/api-electron/src/preload/window-interface.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  /**
+   * Get the minimum size of the window.
+   * @returns {Promise<Number[]>} A promise that resolves to an array containing the minimum width and height of the window.
+   */
+  getMinimumSize(): Promise<ReadonlyArray<number>>;
+
+  /**
+   * Get the parent of the window. Null will be returned if the window has no parent.
+   * @returns {object} The parent window.
```
BenLambertNcl: Type is wrong with the change to the method

packages/api-electron/src/preload/window-interface.ts
```diff
@@ -0,0 +1,296 @@
...
+  /**
+   * Removes all listeners from all window events.
+   */
+  removeAllListeners(): void;
+
+  /**
+   * Send a message to the window.
+   * @param {string|object} message - The message to send to the window. Can be any serializable object.
+   */
+  postMessage(message: string): void;
```
BenLambertNcl: This can be a string or JS Object

packages/api-electron/src/preload/window-interface.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  /**
+ * Gets the id of the current window.
+ * @returns {string} The window id.
+ * @static
+ */
+  static getCurrentWindowId(): number;
```
BenLambertNcl: This returns a string

packages/api-electron/src/preload/window-interface.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  /**
+ * Gets the current window object.
+ * @param {function} callback - Function that is called when the window is created successfully.
+ * @param {function} errorCallback - Function that is called when the window could not be created.
+ * @returns {Window} The window.
+ * @static
+ */
+  static getCurrentWindow(callback: Function, errorCallback: Function): any;
```
BenLambertNcl: Could this not return type `Window`, or will typescript complain?

BenLambertNcl (commented):
Good Work :)

BenLambertNcl (approved):

PR #117, opened 2017-05-15
Fix electron setIcon method

### LINKED ISSUE #103: setIcon is inconsistent between containers
The `setIcon` window function takes a filepath/NativeImage in Electron and an image URL in OpenFin. Changing the Electron version to match OpenFin should be simple, as a NativeImage can be created from a URL, which we can then pass in.

BenLambertNcl:
`setIcon` now takes a URL in Electron. Only way to do this was to use the `request` node module via the `remote` module. Closes #103


packages/api-electron/src/preload/window.ts
```diff
@@ -145,7 +146,19 @@ class Window implements SSFWindow {
   }
 
   setIcon(icon) {
-    return this.asPromise<void>(this.innerWindow.setIcon, icon);
+    const req = request.defaults({ encoding: null });
+    return new Promise<void>((resolve, reject) => {
+        req.get(icon, (err, res, body) => {
+          const image = nativeImage.createFromBuffer(Buffer.from(body));
+          if (image.isEmpty()) {
+            reject(new Error('Image could not be created from the URL'));
```
AndyMJones, line 163: Currently only electron throws an error when it cannot load the image. Probably worth adding an issue to look into OpenFin throwing this error too.

AndyMJones (approved):

PR #118, opened 2017-05-15
containerjs => ContainerJS

ColinEberhardt:
Updates case-sensitive URLs as a result of repo name having changed from containerjs to ContainerJS


BenLambertNcl (approved):
LGTM, just needs rebasing :)

PR #130, opened 2017-05-17
Fixed OpenFin default position

AndyMJones:


packages/api-tests/test/window.spec.js
```diff
@@ -868,5 +863,21 @@ describe('Window API', function(done) {
 
       return chainPromises(steps);
     });
+
+    it.skip('Should be created in the center if no x and y #ssf.Window(center)', function() {
```
BenLambertNcl: Should this still be skipped?

packages/api-tests/test/window.spec.js
```diff
@@ -868,5 +863,21 @@ describe('Window API', function(done) {
...
+      const windowTitle = 'windownamecenter';
+      const windowOptions = getWindowOptions({
+        name: windowTitle,
+        center: false
+      });
+
+      const steps = [
+        ...setupWindowSteps(windowOptions),
+        () => callAsyncWindowMethod('getPosition'),
+        (result) => assert.deepEqual(result.value, [560, 220])
```
BenLambertNcl: Im guessing this is not the center of the screen for the CI environment? You can set the resolution via powershell [like so](https://github.com/FeodorFitsner/selenium-tests/blob/screen-resolution/appveyor.yml#L3-L4)

BenLambertNcl (approved):

PR #136, opened 2017-05-19
Auto generate documentation from interface and test output

BenLambertNcl:
Allows generation of the documentation from the typescript interface and the test results. This will eventually be done on CI and pushed to `gh-pages`, but as this PR is already very large I thought I would leave it for a separate PR, unless anyone wants me to add it onto this. Most of the changes come from the large changes in the docs folder, not form the actual code :)

image


packages/api-specification/generate-docs.js
```diff
@@ -0,0 +1,224 @@
...
+  10: 'red',
+  25: 'yellow',
+  50: 'yellowgreen',
+  75: 'green',
+  100: 'brightgreen'
+};
+
+const getBadgeColor = (percent) => {
+  let colorCode = '';
+  Object.keys(badgeColors).some((key) => {
```
AndyMJones: I see why you are using `some()` but I think it would be more readable if you used a filter or a reduce instead so that you dont need to set `colorCode` and can just use the returned value.

packages/api-specification/generate-docs.js
```diff
@@ -0,0 +1,224 @@
...
+  mdObject[classObj.name] = `# ${classObj.name}  \n`;
+  if (classObj.children) {
+    classObj.children.forEach((child) => {
+      render(child, classObj.name);
+    });
+  }
+};
+
+const generateMethod = (methodObj, className) => {
+  let methodData = '';
```
AndyMJones: You could just define this as `const methodData = methodObj.flags.isStatic ? ' (static)' : '';`

packages/api-specification/generate-docs.js
```diff
@@ -0,0 +1,224 @@
...
+  }
+};
+
+const getParamList = (paramList) => {
+  let params = [];
+  if (!paramList) {
+    return params;
+  }
+
+  paramList.forEach((arg, i) => {
```
AndyMJones: You dont need the `i`.

AndyMJones (changes requested):
Just some small changes but overall looks good!

Good job :+1:

ColinEberhardt:
@BenLambertNcl is there anywhere I can see the rendered output.

BenLambertNcl:
@ColinEberhardt I have updated the description to include a screenshot of the window API

AndyMJones (approved):

packages/api-electron/src/preload/app.ts
```diff
@@ -1,5 +1,7 @@
-const ready = () => Promise.resolve();
+class app implements ssf.App {
```
jleft, line 1: Why change this to a class? Seems like a unrelated change and I'm not sure what the benefit is? If it does make sense, the should `app` be capitalised to `App` for consistency?

packages/api-specification/interface.ts
```diff
@@ -0,0 +1,323 @@
+declare namespace ssf {
```
jleft, line 1: Will this work if the API isn't accessed from global scope (on the `ssf` global)? I.e. through an import?

BenLambertNcl:
@jleft I will change the typescript definition in a new PR, as this is currently blocking other work from happening.

PR #140, opened 2017-05-22
Added specific OpenFin and Electron tests

AndyMJones:
These tests could be done as a single one with just the expected answer changed, however i thought it would be a good idea to put in an example of how to add specific tests for each container but also show that the test matrix still works with unit tests like these.

This way of writing specific tests instead of using tags means that its simpler to run tests overall however each describe that needs to do this will need to have an if statement in it.

### DIFF
--- a/packages/api-tests/test/window.spec.js
+++ b/packages/api-tests/test/window.spec.js
@@ -179,22 +179,6 @@ describe('Window API', function(done) {
       return chainPromises(steps);
     });
 
-    it.skip('Should return the id of the window #ssf.Window.getId', function() {
-      const windowTitle = 'windownameclose';
-      const windowOptions = getWindowOptions({
-        name: windowTitle
-      });
-
-      const steps = [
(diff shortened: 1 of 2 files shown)


packages/api-tests/test/window.spec.js
```diff
@@ -887,6 +871,40 @@ describe('Window API', function(done) {
...
+        const steps = [
+          ...setupWindowSteps(windowOptions),
+          () => selectWindow(app.client, 1),
+          () => callWindowMethod('getId'),
+          (result) => assert.equal(result.value, 2)
+        ];
+
+        return chainPromises(steps);
+      });
+    } else {
```
BenLambertNcl: I would change this to `if (process.env.MOCHA_CONTAINER === 'openfin')`. I think it reads a little better, as well as being extensible if we add another container at some point.

BenLambertNcl (approved):

PR #143, opened 2017-05-23
Add event tests

AndyMJones:


packages/api-tests/test/window.spec.js
```diff
@@ -1038,4 +1041,252 @@ describe('Window API', function(done) {
...
+        callback(window.listenEventResult);
+      };
+      /* eslint-enable no-undef */
+      return executeAsyncJavascript(app.client, script);
+    };
+
+    const retrieveDelay = 500;
+
+    it.skip('Should emit the blur event #ssf.WindowEvent.blur', function() {
+      // Electron works but OpenFin doesnt seem to set the value
```
BenLambertNcl, line 1068: If this is a difference in how the containers work then I think we should add it as an issue.

BenLambertNcl (approved):
This file is getting pretty huge, would be nice if we could split it up somehow

packages/api-tests/test/window.spec.js
AndyMJones: Unsure if its a difference between how the containers work or if our implementation has broken it as it looks correct in their APIs. I'll add an issue either way.

PR #146, opened 2017-05-24
Change browser API to use typescript

### LINKED ISSUE #124: Change browser API to use typescript
The Browser API does not currently use typescript. This should be changed to be consistent with the other 2 APIs

BenLambertNcl:
Currently, the browser window API does not implement the interface. This will be changed once the typescript interface has been changed into `core` and `extended` functionality.

Closes #124


AndyMJones (approved):
Need to add a git ignore for the iterface but other than that looks good

PR #159, opened 2017-05-25
Added WindowOptions type

ColinEberhardt:
I could do with a bit of help documenting the properties with TODOs. They seem quite obvious, but I can't find the electron equivalents.


packages/api-specification/interface.ts
```diff
@@ -28,7 +127,7 @@ declare namespace ssf {
      * @param {boolean} flag - Flag to start or stop the window flashing.
      * @returns {Promise<void>} A promise which resolves to nothing when the function has completed.
      */
-    flashFrame(flag: boolean): Promise<void>;
+    flashFrame(flag: Number): Promise<void>;
```
BenLambertNcl: Why has this changed?

packages/api-specification/generate-docs.js
```diff
@@ -70,6 +70,12 @@ const generateInterface = (interfaceObj) => {
 const generateProperty = (prop, className) => {
   const typeArgsString = getType(prop.type);
   mdObject[className] += `**${prop.name}${prop.flags.isOptional ? '?' : ''}**: \`${typeArgsString}\`  \n`;
+  const comment = prop.comment;
+  if (comment) {
+    if (comment.shortText) {
```
BenLambertNcl, line 75: Could we do `if (comment && comment.shortText)`, or are there other fields we might want in the future?

BenLambertNcl (commented):
Looks really good. @ColinEberhardt Do you want me to pull this down and add the descriptions once those comments have been addressed?

packages/api-specification/interface.ts
ColinEberhardt: Ooops! thanks for spotting

packages/api-specification/generate-docs.js
ColinEberhardt: Yes, although I've got an idea that I think could make this transformation a bit simpler, so I'll give that a go first :-)

ColinEberhardt:
> Looks really good. @ColinEberhardt Do you want me to pull this down and add the descriptions once those comments have been addressed?

Thanks @BenLambertNcl that would be good.

BenLambertNcl:
@ColinEberhardt I have updated the TODOs

ColinEberhardt:
> @ColinEberhardt I have updated the TODOs

Thanks @BenLambertNcl  - I've rebased, so once tests pass, should be good to merge.

AndyMJones (approved):

BenLambertNcl (approved):

PR #160, opened 2017-05-25
remove the interface copy step

ColinEberhardt:
As a general observation, ContainerJS has quite a few postinstall steps that copy files around. These can make things pretty confusing! 

This eliminates one of those steps, by configuring TypeScript to look in the `node_modules/containerjs-api-specification` folder.


BenLambertNcl (approved):
Much nicer :+1:

PR #165, opened 2017-05-25
A simpler approach to generating the docs

ColinEberhardt:
I've been exploring a different approach to generating the docs ...

A few random thoughts:
 - transforming between documents with different syntax(e.g. HTML=> TXT, JSON => MD) is always a pain. 
 - the API docs are conceptually nested and tree-like, flattening them to markdown loses this structure

Anyhow, this approach is based on a concept I played around with a while back, creating an XSLT-style transformation language for JSON http://blog.scottlogic.com/2016/06/22/xslt-inspired-ast-transforms.html

This modified approach uses the `json-transforms` package I created for transforming the typedoc output into a JSON description of a HTML document. It then uses the `html2json` package to perform this transform. Finally `gray-matter` adds the front matter.

What do you think? The pattern matching approach should make it easier to re-use transformations. Also, I think it is a little easier to follow the transformation process itself.


BenLambertNcl:
Looks like a great start. I pulled it down to have a look at the output. One thing that would be nice would be to have the promise types. In the output, the return types for window methods are `Promise` whereas it would be nice to get `Promise<void>` or `Promise<ReadOnlyArray<number>>` for example. This information is given in `typeArguments`:
```javascript
"signatures": [
	{
		"id": 149,
		"name": "ready",
		"kind": 4096,
		"kindString": "Call signature",
		"flags": {},
		"type": {
			"type": "reference",
			"name": "Promise",
			"typeArguments": [
				{
					"type": "intrinsic",
					"name": "any"
				}
			]
		}
	}
],
```
Another idea might be to add classes to certain parts of the output (e.g. the return type `<p>` might want `class=code`). I guess this is pretty simple with the way the html is now generated :).

Looks much cleaner than before though. Will work well as a single page too as it seems we were heading more towards that :+1:

packages/api-specification/transform-type-info.js
```diff
@@ -0,0 +1,106 @@
...
+    '.{.kindString === "Class"}', d =>
+      section([
+        h1(d.match.name),
+        section(d.runner(d.match.children))
+      ])
+  ),
+  jsont.pathRule(
+    '.{.kindString === "Method"}', d =>
+      // we take just the first signature as the definition of the method, not
+      // sure when there might be multiple signatures?!
```
BenLambertNcl: I wasn't sure either. It seems like Typescript supports multiple method signatures, but only 1 method implementation. E.g.
```javascript
aMethod(a: string);
aMethod(a: number);
aMethod(a: any) {
  console.log(a);
}
```
So it seems like a very specialized use case. The same could be achieved with union types I think

packages/api-specification/transform-type-info.js
```diff
@@ -0,0 +1,106 @@
...
+
+// html elements are a bit complex, this simplifies their construction
+const element = (tag, child) => ({
+  node: 'element',
+  tag,
+  // if child is not an array, turn it into one - also, filter out any undefined children
+  child: (Array.isArray(child) ? child : [child]).filter(d => d)
+});
+
+// text elements have not children
```
BenLambertNcl, line 28: not => no

AndyMJones:
This is looking really good, much easier to understand :+1:

ColinEberhardt:
I think this is good to merge - the generated HTML output is generally OK, but not styled at all.

A few other changes

 - I've removed another copy step (yay!), the script that generates the docs has an output-folder option. I've tried to make it more like a CLI tool
 - I've moved some `docs` specific `.gitignore` into the subfolder

BenLambertNcl (approved):

PR #171, opened 2017-05-31
Feature/api definitions

kirilpopov:
This is an attempt to define an API to unify what we currently have in ContainerJS, [the working group APIs](https://symphonyoss.atlassian.net/wiki/display/WGDWAPI/Proposed+Standard+API+Specifications)
and things that we at Tick42 think are useful based on our experience.

This is an initial version that aims to start a discussion - some parts of it are not fully defined (e.g. window events) and some
upcoming parts are just mentioned in comments (e.g. interop). 

This also contains a [document](https://github.com/kirilpopov/ContainerJS/blob/feature/api-definitions/packages/api-specification/api/confluence.md) describing how to map the [the working group APIs in confluence (https://symphonyoss.atlassian.net/wiki/display/WGDWAPI/Proposed+Standard+API+Specifications) to the current API.

The starting point when reading the code should be [ssf.ts](https://github.com/kirilpopov/ContainerJS/blob/feature/api-definitions/packages/api-specification/api/ssf.ts) file that describes the ssf root object as exposed to clients applications.


packages/api-specification/api/windows.ts
```diff
@@ -0,0 +1,313 @@
...
+    activate(focus?: boolean): Promise<void>;
+
+    /**
+     * Captures an image of the window
+     * @param options
+     */
+    capture(options?: CaptureOptions): Promise<Base64ImageData>;
+
+    /** Get the bounds of the window */
+    getBounds(): Bounds;
```
BenLambertNcl, line 229: I think all window methods should be asynchronous, as OpenFin methods are all async.

packages/api-specification/api/windows.ts
```diff
@@ -0,0 +1,313 @@
...
+
+/**
+ * A window running in the container.
+ * It implements EventEmitter which means users can subscribe for events using on, once , addEventListener etc.
+ * The list of events is in WindowsEvents structure;
+ */
+export interface Window extends EventEmitter {
+
+    /** Get or set the max width of the window. Might be undefined if no restrictions applied**/
+    maxWidth?: number;
```
BenLambertNcl, line 150: I think these will need to be getters/setters as they can change after the window is open. OpenFin's methods are also async, see below.

packages/api-specification/api/common.ts
```diff
@@ -0,0 +1,20 @@
+// contains common types
+export interface Bounds {
+    top: number;
```
BenLambertNcl, line 3: In Electron, `top` and `left` are `x` and `y`. If we are trying to stick to the Electron API as much as possible then maybe this should be changed?

packages/api-specification/api/windows.ts
```diff
@@ -0,0 +1,313 @@
...
+     * Minimize the window.
+     * @returns {Promise<void>} A promise that resolves to nothing when the window has minimized.
+     */
+    minimize(): Promise<void>;
+
+    /**
+     * Send a message to the window.
+     * @param {string|object} message - The message to send to the window. Can be any serializable object.
+     */
+    sendMessage(message: string | object): Promise<void>;
```
BenLambertNcl, line 303: The HTML5 API is [`postMessage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) so we should maybe use that?

packages/api-specification/api/windows.ts
```diff
@@ -0,0 +1,313 @@
...
+    /**
+     * The window that is focused in this application, otherwise returns null
+     */
+    focusedWindow: Window;
+
+    /**
+     getById(windowId: string): Window;
+
+     /** Opens a new window */
+    open(options: WindowOptions): Promise<Window>;
```
BenLambertNcl, line 22: I prefer using the constructor pattern that is used in some of the newer browser APIs (like notifications). Might be better to keep it like this though if `window.open` is used in the symphony application.

packages/api-specification/api/windows.ts
```diff
@@ -0,0 +1,313 @@
+import {Bounds, UnsubscribeFunction} from "./common";
+import {Base64ImageData, CaptureOptions} from "./system";
+import EventEmitter = NodeJS.EventEmitter;
+
+export interface WindowsAPI {
+
+    /** Returns the current window */
+    current: Window;
```
BenLambertNcl, line 8: If we have `getById()`, then we might want to have these as `getCurrent()` and `getFocusedWindow()` etc. this would match both Electron + OpenFinas well.

packages/api-specification/api/application.ts
```diff
@@ -0,0 +1,33 @@
+export interface ApplicationAPI {
```
BenLambertNcl, line 1: We might need to clarify the meaning of `Application` for OpenFin, as currently OpenFin windows that are not child windows are created using `fin.desktop.Application()`. Would this API be used for each OpenFin application, or ContainerJS application? If we want this per ContainerJS application, then we might need to rethink the `setBadgeCount()` or the window creation implementation as each ContainerJS application could have multiple top level windows, which could have different badge counts.

BenLambertNcl (commented):
Looks like a great start!

I have left a few comments, mostly high level, but i'm not sure how much leeway we have in changing some of these APIs. I will do another code review once we have finalized on the design :-).

packages/api-specification/api/application.ts
ColinEberhardt: I guess a wider question is, should the API support multiple applications?

With OpenFin you can create a [new Application instance](http://cdn.openfin.co/jsdocs/stable/fin.desktop.Application.html#Application) through the JavaScript API.

Personally I think this is a bit of an edge case. I can't see when you'd want to do this in practice? I'd be happy with ContainerJS treating the Application as a singleton.

packages/api-specification/api/ssf.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Messaging API - communicate with other windows
+     */
+    messages: MessagesAPI;
+
+    /**
+     * Notifications API - publish notifications to the user
+     */
+    notifications: NotificationsAPI;
```
ColinEberhardt, line 30: I guess the interesting thing here is that both within the browser and Electron container users will see two notification APIs:

```
new Notification(...)
new ssf.Notification(...)
```

This is probably unavoidable, but something that users need to be warned about.

packages/api-specification/api/system.ts
```diff
@@ -0,0 +1,88 @@
...
+     * @param options
+     */
+    captureAllWindows: (options?: CaptureOptions) => Promise<WindowImageData>;
+
+    /**
+     * Whenever user activity occurs (keyboard or mouse input) the callback will be invoked.
+     * If additional activity occurs within the period given by the 'throttle', then the callback will be
+     * invoked at time: X + throttle (where X is the time of last activity report send).
+     */
+    onUserActivity(callback: ()=> void, throtle: number);
```
ColinEberhardt, line 35: throt**t**le

packages/api-specification/api/windows.ts
ColinEberhardt: They are, although I do wonder whether we should wrap them to be synchronous? All the equivalent Electron methods are synchronous.

I wonder whether this is an OpenFin implementation detail that has 'leaked' into the public API?

packages/api-specification/api/windows.ts
```diff
@@ -0,0 +1,313 @@
...
+     getById(windowId: string): Window;
+
+     /** Opens a new window */
+    open(options: WindowOptions): Promise<Window>;
+
+    /**
+     * Receive notifications when a new window is opened
+     * @returns Function Execute the function to unsubscribe
+     */
+    onWindowOpened(callback: (window: Window) => void): UnsubscribeFunction;
```
ColinEberhardt, line 28: Is there any reason why `Window` is an event emitter, whereas the `WindowsAPI` has explicit methods for each event type?

I'd prefer to stick to the event emitter pattern!

packages/api-specification/api/windows.ts
```diff
@@ -0,0 +1,313 @@
...
+    /** Returns all windows in the application */
+    all: Window[];
+
+    /**
+     * The window that is focused in this application, otherwise returns null
+     */
+    focusedWindow: Window;
+
+    /**
+     getById(windowId: string): Window;
```
BenLambertNcl, line 19: Missing closing comment

packages/api-specification/api/windows.ts
kirilpopov: Yes these should be get* methods . Will update

packages/api-specification/api/windows.ts
kirilpopov: Yes these can be implemented as properties with getter  and setters. (I prefer this over get* set* methods),

```javascript
window.maxWidth = 10;
console.log(window.maxWidth);
```

What do you think?

BenLambertNcl: Not sure it would work, unless we have them as read only and use them instead of get methods. If we had an open window and did `window.maxWidth = 10;`, then we need a way of calling the underlying method to set the value. I think the only way to do this would be to watch for variable changes? Unless you can think of another way? :-)

packages/api-specification/api/windows.ts
kirilpopov: Usually these are sync operations , not sure why openfin did it that way. 
Also I think usage is better

```
//move my window with 10px
window.setBounds({left: window.getBounds().left + 20});

//vs
window.getBounds().then((bounds)=>{
 window.setBounds({left: bounds.left + 20});
});

File: packages/api-specification/api/windows.ts (hunk 4 above)
Comment (kirilpopov) [reply packages/api-specification/api/windows.ts:303]: Ok will update

File: packages/api-specification/api/application.ts (hunk 7 above)
Comment (kirilpopov) [reply packages/api-specification/api/application.ts:1]: I didn't know that. 
We should decide if we want to have access to other applications and be able to start some of them (however I'm not sure how other apps will be discovered in Electron case)

File: packages/api-specification/api/system.ts (hunk 9 above)
Comment (kirilpopov) [reply packages/api-specification/api/system.ts:35]: thanks will change

File: packages/api-specification/api/windows.ts (hunk 11 above)
Comment (kirilpopov) [reply packages/api-specification/api/windows.ts:19]: thanks - will update

File: packages/api-specification/api/windows.ts (hunk 10 above)
Comment (kirilpopov) [reply packages/api-specification/api/windows.ts:28]: Ok will change

File: packages/api-specification/api/windows.ts (hunk 2 above)
Comment (kirilpopov) [reply packages/api-specification/api/windows.ts:150]: I was thinking of using properties, something like (in typescript)

```
class Window{
    private _maxWidth: number;

    get maxWidth():boolean {
        return this._maxWidth;
    }
    set maxWidth(width:number) {
        // do some container specific magic
        this._maxWidth= width;
    }
}
````

Comment (BenLambertNcl) [reply packages/api-specification/api/windows.ts:150]: Ah yeah I forgot we were using TypeScript :smile:. Sounds like a good idea to me.

File: packages/api-specification/api/application.ts (hunk 7 above)
Comment (BenLambertNcl) [reply packages/api-specification/api/application.ts:1]: I think we can keep it as is for the time being, and have this as a single ContainerJS application. We can always change it in the future if we hit any problems

File: packages/api-specification/api/notifications.ts [hunk 12]
```diff
@@ -0,0 +1,70 @@
...
+
+export interface NotificationAction {
+    action: string;
+    title: string;
+    icon?: string;
+}
+
+declare class Notification extends EventTarget {
+    constructor(title: string, options?: NotificationOptions);
+
```
lneir, line 48: looks like missing requestPermission method.

packages/api-specification/api/notifications.ts
```diff
@@ -0,0 +1,70 @@
...
+    onclick: EventListenerOrEventListenerObject;
+    onerror: EventListenerOrEventListenerObject;
+
+    close(): void;
+
+    readonly title: string;
```
lneir, line 54: also need likely need a sticky option

packages/api-specification/api/notifications.ts
```diff
@@ -0,0 +1,70 @@
...
+    readonly dir: NotificationDirection;
```
lneir, line 55: need to think how we might handle non-std fields.  e.g., blink and blink color.

packages/api-specification/api/application.ts
lneir: in general need to think about which aspects are mandatory and optional

packages/api-specification/api/windows.ts
```diff
@@ -0,0 +1,313 @@
...
+     */
+    transparent?: boolean;
+}
+
+/**
+ * A window running in the container.
+ * It implements EventEmitter which means users can subscribe for events using on, once , addEventListener etc.
+ * The list of events is in WindowsEvents structure;
+ */
+export interface Window extends EventEmitter {
```
lneir, line 147: one issue i have here is that is kind of re-inventing the wheel.  javascript already has a Window object and window.open nicely returns one these.  so for javascript already using window.open this might require significant refactoring and might not be able to achieve same functionality. for example, window.opener.  here the equivalent looks like window. getParentWindow() but it still don't think will give same functionality.  what about postMesage and 'closed'

for example, one use case we have is we want child window to able to share large object (redux state) with parent.  with window.opener this is possible.  not sure here.

is it possible to augment existing window object (or inherit from it) and add missing/optional things we need here???

packages/api-specification/api/system.ts
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Info about all displays available
+     */
+    displays: () => Display[];
+
+    /**
+     * Capture
+     * @param options
+     */
+    capture: (options?: CaptureOptions) => Promise<Base64ImageData>;
```
lneir, line 13: how is this capture working?  would like to see something similar to ScreenSnippet here https://symphonyoss.atlassian.net/wiki/display/WGDWAPI/ScreenSnippet+API

packages/api-specification/api/system.ts
```diff
@@ -0,0 +1,88 @@
+import {Bounds, Size} from "./common";
+
+export interface SystemAPI {
+    /**
+     * Info about all displays available
```
lneir, line 5: other interface needed is for screen sharing... https://symphonyoss.atlassian.net/wiki/display/WGDWAPI/getMediaSources+API which is essentially equal to https://electron.atom.io/docs/api/desktop-capturer/#desktopcapturer

packages/api-specification/api/messages.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Currently this is based on OpenFin InterApplicationBus - http://cdn.openfin.co/jsdocs/stable/fin.desktop.InterApplicationBus.html
+ *
+ * TBD - Request/response and streaming
+ */
+export interface MessagesAPI{
+    send(windowId: string, topic :string, message: string|object): void;
```
lneir, line 7: what is windowId here?

packages/api-specification/api/windows.ts
lneir: yes

lneir, line 303: how to receive a message?  ??? win.addEventListener('message', callback);

packages/api-specification/api/ssf.ts
```diff
@@ -0,0 +1,46 @@
...
+export interface SSF {
+    /**
+     * Window management API - create, discover and manage windows
+     */
+    windows: WindowsAPI;
+
+    /**
+     * Messaging API - communicate with other windows
+     */
+    messages: MessagesAPI;
```
lneir, line 25: how is this diff than window.sendMessage (postMesage) below?

packages/api-specification/api/windows.ts
kirilpopov: It's not about reinventing but for defining a common contract for container window. Note that this is not just about the current window, it's a way to address any window in your application from the current one (even if they don't have a parent/child relationship). 

Any container I've worked with (Electron, OpenFin, GlueDesktop (which is Tick42 CEF based container)) has similar contract for windows, because the standard browser windowing API is not good enough for desktop applications.

Augmenting the existing window is actually a proposal for implementation of this contract.

packages/api-specification/api/notifications.ts
kirilpopov: is this a standard field?

packages/api-specification/api/windows.ts
ColinEberhardt: > for example, one use case we have is we want child window to able to share large object (redux state) with parent. with window.opener this is possible. not sure here.

One problem you'll encounter here is that the process model differs between various containers / browser.

Within a browser, a window opened via window.open is in the same process, therefore window.opener from the child process uses the same JavaScript context, and as a result, sharing objects, such as redux state becomes possible.

This is also true within OpenFin.

However, in Electron each renderer window has its own process, so you cannot just use window.opener to access the parent context and state.

This is why projects such as https://github.com/samiskin/redux-electron-store exist.

Within ContainerJS we are looking to create a consistent process model across all containers, as well as a consistent API.

lneir: using window.open (with sandbox on) in electron like this works just fine, the child can use the window.opener.  we have this working currently in electron.  when window.open url has the same domain, then it runs in the same process and the child can access the parent.  

but yes if in diff process then window.opener won't work without some extra method to proxy between process across ipc.

ColinEberhardt: Yes, we recently discovered this behaviour, and @BenLambertNcl is exploring it further via #195

PR #173, opened 2017-06-01
Wip Add api bundle

AndyMJones:
Currently this is a working package that you can include to run in Browser and OpenFin. It also doesn't break Electron however you still need to globally install the Electron CLI.

This has been tested with notifications and a few other methods which all seem to be working as intended however it probably needs extensive testing.

### DIFF
--- a/packages/api-openfin/package.json
+++ b/packages/api-openfin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "containerjs-api-openfin",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "description": "",
   "main": "build/lib/index.js",
   "module": "build/es/index.js",
--- a/packages/api-openfin/src/shims.ts
+++ b/packages/api-openfin/src/shims.ts
@@ -1,16 +1,18 @@
-window.Notification = function(title, options) {
-  if (!options) {
-    options = {};
-  }
+if ('fin' in window) {
+  window.Notification = function(title, options) {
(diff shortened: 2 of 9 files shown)


packages/api-openfin/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "containerjs-api-openfin",
-  "version": "0.0.1",
+  "version": "0.0.2",
```
AndyMJones, line 3: Had to update the version for npm to push the changes to the build folder

packages/api-openfin/src/shims.ts
```diff
@@ -1,16 +1,22 @@
-window.Notification = function(title, options) {
-  if (!options) {
-    options = {};
-  }
+interface Window {
```
AndyMJones: I don't like having this one interface in this file when all the others are in their own. However this is only needed for OpenFin and without it it throws the error that window doesn't have a fin property and without window the application throws an error when running in browser and Electron.

ColinEberhardt: It's probably worth moving this comment, or something similar, into the `shim.ts` file itself.

ColinEberhardt (approved):
This all looks good to me - thanks :-)

Can you please updated the `README` in the root of ContainerJS to reference this bundle?

BenLambertNcl (approved):
LGTM!

PR #174, opened 2017-06-01
Add tests for browser API

BenLambertNcl:
Uses the current tests to test the browser API, implemented using webdriverio and selenium-standalone.  The `WindowCore` interface has also been added to include all of the browser specific methods.


packages/api-specification/generate-docs.js
```diff
@@ -52,12 +56,13 @@ const generateMethod = (methodObj, className) => {
...
     if (testMatrixJson && testMatrixJson[`ssf.${className}.${methodObj.name}`]) {
       const testResults = testMatrixJson[`ssf.${className}.${methodObj.name}`];
       const electronBadgeColor = testResults.electron.total > 0 ? getBadgeColor((testResults.electron.passed / testResults.electron.total) * 100) : 'lightgrey';
       const openfinBadgeColor = testResults.openfin.total > 0 ? getBadgeColor((testResults.openfin.passed / testResults.openfin.total) * 100) : 'lightgrey';
-      badgeString = `![${methodObj.name}](https://img.shields.io/badge/Electron-${testResults.electron.passed}%2F${testResults.electron.total}-${electronBadgeColor}.svg) ![${methodObj.name}](https://img.shields.io/badge/OpenFin-${testResults.openfin.passed}%2F${testResults.openfin.total}-${openfinBadgeColor}.svg)`;
+      const browserBadgeColor = testResults.browser.total > 0 ? getBadgeColor((testResults.browser.passed / testResults.browser.total) * 100) : 'lightgrey';
```
jleft: This line is repeated (with browser/openfin/electron varying) several times - might be good to refactor into a function to keep it dry? Could also call it from `getShieldMarkdown` to further reduce repetition.

packages/api-specification/interface.ts
```diff
@@ -139,19 +138,12 @@ declare namespace ssf {
     blur(): Promise<void>;
 
     /**
-     * Closes the window.
+     * Closes the window. Only works on new windows.
```
jleft: What is a 'new' window? A window created using the ContainerJS API? Is this only in browser?

packages/api-tests/package.json
```diff
@@ -4,24 +4,31 @@
...
-    "spectron": "^3.6.2"
+    "spectron": "^3.6.2",
+    "webdriverio": "^4.8.0"
   },
   "devDependencies": {
     "copyfiles": "^1.2.0",
     "cross-env": "^4.0.0",
-    "rimraf": "^2.6.1"
+    "karma": "^1.7.0",
+    "karma-mocha": "^1.3.0",
```
jleft: Are all of these dependencies being used (can't see karma being used)?

packages/api-tests/test/browser-test-setup.js
```diff
@@ -0,0 +1,70 @@
+const webdriverio = require('webdriverio');
+const selenium = require('selenium-standalone');
+
```
jleft, line 3: Looks good, nice to see that the current specs can be run in the browser without much setup 👍 

Will the browser tests run on CI?

packages/api-tests/test/browser-test-setup.js
```diff
@@ -0,0 +1,70 @@
...
+  desiredCapabilities: {
+    browserName: 'chrome',
+    chromeOptions: {
+      // Disables "Chrome is being controlled by testing software" banner.
+      args: ['disable-infobars']
+    }
+  }
+};
+
+// This is designed to replicate the Spectron Application object, apart from launching chrome instead of Electron.
```
jleft, line 21: Just curious, what is the reason for this? Is it just for consistency?

packages/api-tests/test/browser-test-setup.js
```diff
@@ -0,0 +1,70 @@
...
+        } else {
+          selenium.start((err, child) => {
+            if (err) {
+              this.running = false;
+              reject(err);
+            } else {
+              this.running = true;
+              this.driver = child;
+              // Start chrome at index.html
+              this.client = webdriverio.remote(options).init().url('localhost:5000/index.html');
```
jleft: Should the protocol - `http://` - be included?

packages/api-tests/test/window.spec.js
```diff
@@ -112,7 +114,7 @@ describe('Window API', function(done) {
 
     });
 
-    it('Should get the bounds of the window #ssf.Window.getBounds', function() {
+    it('Should get the bounds of the window @no-browser #ssf.Window.getBounds', function() {
```
jleft: Nitpick: adding `@no-browser` means that there are now two ways of conditionally running the tests: using the `MOCHA_CONTAINER` env var, or using this (`@no-browser`) string in the test description, I'm not sure if there's a preference, but if we do stick if with the two different ways then maybe the window core tests can be grouped in a `describe` and all other tests can be grouped in a separate `describe` with `@no-browser` added? Or there could be separate spec files for the core and non-core window methods and these could be run based on the `MOCHA_CONTAINER` env var, which could simplify setup and reduce churn?

packages/api-tests/test/window.spec.js
```diff
@@ -908,6 +904,24 @@ describe('Window API', function(done) {
         return chainPromises(steps);
       });
     }
+
+    if (process.env.MOCHA_CONTAINER === 'browser') {
```
jleft: Nitpick: see comment above on `MOCHA_CONTAINER` vs `@no-browser` - should we be more consistent?

jleft (commented):
Looks really good to me 👍  nice to see that the existing specs can be run in the browser without much change! Just a few questions/comments.

packages/api-specification/interface.ts
BenLambertNcl: Yeah ill clarify that :+1:

packages/api-tests/package.json
BenLambertNcl: Oops! Good spot!

packages/api-tests/test/browser-test-setup.js
BenLambertNcl: Yeah they should work the same as the others. Ill make sure they run fine on my CI before merging

packages/api-tests/test/browser-test-setup.js
BenLambertNcl: So we don't need to change the tests. We use `app.start()`, `app.stop()` etc in the beforeEach and afterEach methods

packages/api-tests/test/window.spec.js
BenLambertNcl: Adding the window core methods into a separate file sounds like a better idea. It will also reduce the size of the test file we have at the moment too.

BenLambertNcl:
Some messaging tests fail intermittently in browser, so these have been stopped from running for now so the fixes to the OpenFin tests, as well as the browser window tests can be merged. PR is ready for review :-)

packages/api-tests/test/test-helpers.js
```diff
@@ -12,8 +12,9 @@ const selectWindow = (client, handle) => {
 const openNewWindow = (client, options) => {
   const script = (options, callback) => {
     ssf.app.ready().then(() => {
-      new ssf.Window(options);
-      setTimeout(() => callback(), 500);
+      new ssf.Window(options, () => {
+        setTimeout(() => callback(), 500);
```
jleft: If the window success callback is being used (which I think is a good thing), can the setTimout be removed?

BenLambertNcl: Oops, thought I removed that. :+1:

BenLambertNcl:
@ColinEberhardt @jleft @JohnEz Can someone review this please so it can be merged :-)

packages/api-browser/src/window.ts
```diff
@@ -5,26 +5,19 @@ import {
...
-  return new Promise((resolve, reject) => {
-    if (win) {
-      fn(win);
-      resolve();
-    } else {
-      reject(windowClosedError);
-    }
-  });
-};
+function getCenterCoordinates() {
```
AndyMJones: Maybe use an arrow function instead just to keep with the same style

packages/api-tests/test/browser-test-setup.js
```diff
@@ -0,0 +1,90 @@
...
+    selenium.start((err, child) => {
+      if (err) {
+        this.running = false;
+        reject(err);
+      } else {
+        this.running = true;
+        this.driver = child;
+        // Start chrome at index.html
+        this.client = webdriverio.remote(options).init().url('http://localhost:5000/index.html');
+        this.client.timeouts('script', 60000);
```
AndyMJones: Do we maybe want to pull these out into a variable?

packages/api-tests/test/messaging.spec.js
```diff
@@ -19,191 +19,193 @@ const defaultWindowOptions = {
...
-    }
-  });
+      return app.start();
+    });
 
-  /* eslint-disable no-undef */
-  const setupScript = (id, topic, callback) => {
-    ssf.MessageService.subscribe(id, topic, (message) => {
-      window.testMessage = message;
+    afterEach(function() {
```
AndyMJones: Arrow function for consistency

AndyMJones (approved):

PR #175, opened 2017-06-01
Fix OpenFin demo not closing

### LINKED ISSUE #172: OpenFin window not closing in demo application
The window created in the OpenFin demo does not close. The window can be force closed (by passing true to `fin.desktop.Window.getCurrent().close(true)`) and the window closes fine without forcing after that. This could be something to do with the `close-requested` event.

BenLambertNcl:
Subscribing to the `close-requested` event stops the window from closing, so the user can perform actions before the window closes. Change the demo to listen for `closed` instead. Closes #172.


ColinEberhardt (approved):

jleft:
Looks good as a fix for the demo +1

However, doesn't this indicate an inconsistency in our API, which is the root of the problem? I.e. that when using ContainerJS in OpenFin it isn't possible to close a window when subscribed to it's `close` event, but it is possible in Electron?

BenLambertNcl:
The only way to achieve consistency is either always force close on OpenFin, or to replicate this functionality in electron. Not sure which option is better?

jleft:
> The only way to achieve consistency is either always force close on OpenFin, or to replicate this functionality in electron. Not sure which option is better?

Although both of those options would be consistent, I think it would leave the user unable to prevent a window from closing or perform any operations before the window closes - I think it would be very useful if the user could do this.

Would it be possible to do something based around the [`Event`](https://developer.mozilla.org/en-US/docs/Web/API/Event) interface? I think this would be more consistent with the web and Electron and feels more idiomatic than the approach used by OpenFin.

I'd propose that the behaviour by default would be that the window will close, regardless of whether the `close` listener is added or not (in OpenFin the close would be forced). However, add the option for the user to prevent the default behaviour (i.e. prevent the close using [`event.preventDefault()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)) within their event handler. Our API can then detect if default has been prevented (i.e. [`event.defaultPrevented`](https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented)) and use this to determine whether this window should be closed or not (in Openfin, force close if `event.defaultPrevented === false`, otherwise don't close). What do you think?

BenLambertNcl:
Yeah I think that's a better idea. Ill add it into another PR.

PR #179, opened 2017-06-05
Created single api demo

AndyMJones:
This PR creates a single package that can be used for demoing browser, Openfin and Electron rather than having to manage 3 different packages. It uses the new api bundle package which links to the 3 specific api packages.

The old demo packages still need to be deleted but i thought i'd wait until these changes had be approved.


packages/api-demo/src/bootstrap.min.css
```diff
@@ -0,0 +1,6 @@
+/*!
```
BenLambertNcl, line 1: Should this be gitignored?

packages/api-demo/src/containerjs-bundle.js
```diff
@@ -0,0 +1,4263 @@
+(function (global, factory) {
```
BenLambertNcl: Same about gitignore

BenLambertNcl (commented):
With moving all of the demos into 1, we should be ok to delete `api-specification/src` I think. I would delete the old demos in this PR, but its up to you.

packages/api-demo/src/bootstrap.min.css
AndyMJones: Not in its current state but i could bring it in as an npm package if you think that would be better?

BenLambertNcl: No I think as it is now is better :)

BenLambertNcl (approved):

PR #180, opened 2017-06-07
API demo fixes and removal of redundant projects

ColinEberhardt:
`api-demo`:
- removed dependency on express, using live-server instead
- added a readme
- added missing containerjs-api-electron dependency

I've also removed all the `api-*-demo` packages

The project root README is now focused at developers, with the end-users being directed to the website for the getting started guide.


README.md
```diff
@@ -116,14 +14,17 @@ To get started, run the following from the project root:
...
-npm run openfin
-```
-or refer to the demo projects
+If you want to see ContainerJS in action, the `api-demo` project has a fully-featured demo that can be run against various containers.
+
+The ContainerJS repo contains the following:
+
+ - `api-specification` - the ContainerJS API specified in TypeScript.
+ - `api-browser`, `api-electron`, `api-openfin` - various container-specific implementations of this API.
+ - `api-bundle` - a bundle that contains the client-side code for all of the different ContainerJS implementations. This allows exactly teh same application code to be run against each container.
```
BenLambertNcl: teh => the

packages/api-demo/package.json
```diff
@@ -4,13 +4,13 @@
   "description": "A bundle of containerJS APIs",
   "main": "src/index.js",
   "scripts": {
-    "serve": "node server.js",
+    "serve:no-browser": "live-server --mount=/:src --mount=/scripts/containerjs-api-bundle:node_modules/containerjs-api-bundle/build/containerjs-bundle.js --no-browser",
```
BenLambertNcl: Could this not be `npm run serve -- --no-browser`

packages/api-specification/src/window/window-api-demo.js
```diff
@@ -1,149 +0,0 @@
...
-
-      win.addListener('blur', () => {
-        addListItem('blur');
-      });
-
-      win.addListener('focus', () => {
-        addListItem('focus');
-      });
-
-      win.addListener('closed', () => {
```
BenLambertNcl, line 63: Note that this fixed the OpenFin demo. I think this will need changing in the new demo project

BenLambertNcl (approved):
:+1:

packages/api-demo/package.json
ColinEberhardt: Nice - I had no idea you could do that :+1:

PR #181, opened 2017-06-07
Add styling to documentation pages

BenLambertNcl:
Add new styling to documentation pages. Sample:
image

image


packages/api-specification/transform-type-info.js
```diff
@@ -94,72 +103,82 @@ const documentClass = (className) => {
...
         section([
-          h3(d.match.name),
+          h4(d.match.name, {class: 'method-name', id: `${className}-${d.match.name}`}),
+          d.match.results ? section([
+            p('Electron'),
+            p(`${d.match.results.electron.passed}/${d.match.results.electron.total}`, {class: 'test-result ' + testResultPercentage(d.match.results.electron.passed, d.match.results.electron.total)}),
+            p('OpenFin'),
+            p(`${d.match.results.openfin.passed}/${d.match.results.openfin.total}`, {class: 'test-result ' + testResultPercentage(d.match.results.openfin.passed, d.match.results.openfin.total)}),
+            p('Browser'),
+            p(`${d.match.results.browser.passed}/${d.match.results.browser.total}`, {class: 'test-result ' + testResultPercentage(d.match.results.browser.passed, d.match.results.browser.total)})
```
ColinEberhardt: A few very minor points ...

- I don't think `p` tags make much sense here, how about `h5` for the container name, and a `span` for the results
- There's a bit of duplication on the 3 lines that output the results, might be worth moving into a separate function.

ColinEberhardt (changes requested):
Looks really good to me - the one part I'm unsure of is the need for the extra 'navigation' file. Typically navigation structures are constructed from the page meta-data, via the `pages` Jekyll variable.

BenLambertNcl:
The navigation page is needed because the docs is now outputted as 1 file, so we need a way of being able to generate the menu. We could do this with JavaScript if you think that's a better option?

ColinEberhardt:
> The navigation page is needed because the docs is now outputted as 1 file, so we need a way of being able to generate the menu. We could do this with JavaScript if you think that's a better option?

Good point, and I agree, using JavaScript would be better. I'd avoid making the Jekyll template and configuration any more complicated than it needs to be,

BenLambertNcl:
Updated :-)

packages/api-specification/transform-type-info.js
```diff
@@ -94,72 +107,77 @@ const documentClass = (className) => {
...
-        ])
+        ], {class: 'property', id: d.match.name})
     ),
     jsont.pathRule(
       '.{.kindString === "Call signature" || .kindString === "Constructor signature"}', d =>
         section([
-          h3(d.match.name),
+          h4(d.match.name, {class: 'method-name', id: `${className}-${d.match.name}`}),
+          d.match.results ? section([
+            ...testResults('Electron', d.match.results.electron),
```
ColinEberhardt, line 140: nice use of spread 👍

ColinEberhardt (approved):
thanks for the updates -looks really good.

PR #188, opened 2017-06-09
bumped version number

ColinEberhardt:
Currently `npm i containerjs-api-electron` fails because it depends on `containerjs-api-specification` 0.0.1, which has a failing post install step (yuck!)

This just bumps the version number so that I can release these fixes:

https://github.com/symphonyoss/ContainerJS/pull/185

After this, I am NOT doing any more manual versioning. It's horrible ;-)

### DIFF
--- a/packages/api-demo/package.json
+++ b/packages/api-demo/package.json
@@ -24,8 +24,8 @@
   },
   "homepage": "https://github.com/symphonyoss/ContainerJS#readme",
   "dependencies": {
-    "containerjs-api-bundle": "^0.0.1",
-    "containerjs-api-electron": "^0.0.1",
+    "containerjs-api-bundle": "0.0.2",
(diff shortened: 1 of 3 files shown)


BenLambertNcl (approved):
Could we not add `^` to use the latest version, or would this break something? Or will lerna do the bumping for us when we publish?

ColinEberhardt:
> Could we not add ^ to use the latest version, or would this break something? Or will lerna do the bumping for us when we publish?

It's a good point, and I honestly don't know. For the short term, I just want to make sure ContainerJS is useable by installing via npm :-)

Once it works, then yes, it definitely needs tidying up!

BenLambertNcl:
Just noticed that `packages/api-tests/package.json` also has a reference to electron v0.0.1. Will need to change that as well so tests will run correctly

Issue #191, opened 2017-06-11
Electron TypeScript support (build failing)


jleft:
[TypeScript support in Electron was recently announced](https://electron.atom.io/blog/2017/06/01/typescript).

> Starting at version 1.6.10, every release of Electron includes its own TypeScript definition file. When you install the `electron` package from npm, the `electron.d.ts` file is bundled automatically with the installed package.

> If you were already using third-party definitions like `@types/electron` and `@types/node`, you should remove them from your Electron project to prevent any collisions.

This has lead to the [build failing](
https://travis-ci.org/symphonyoss/ContainerJS/jobs/241676239). We're using `"electron": "^1.6.6"` and `"@types/electron": "^1.4.37"`, so the latest versions of these packages, which includes the changes mentioned above, are being installed.

We can specify the exact package version in `package.json` to fix the build or consider updating to a more recent version. I think it's also worth consider how we're going to test/support different version of Electron/OpenFin etc.

andy-lee-eng:
Now using an up to date version of Electron

PR #192, opened 2017-06-11
Use exact Electron versions

jleft:
Fixes the build. Relates to #191.


ColinEberhardt:
Well spotted! I was wondering what had changed, seeing PR builds that were passing previously, fail when merged.

ColinEberhardt (approved):

PR #190, opened 2017-06-10
Fix AppVeyor test setup to fail build when tests fail

jleft:
Ensure that the AppVeyor  build fails if `npm [run] test`  fails.

Fixes #177.

PR #193, opened 2017-06-11
Update scripts across packages for consistency

### LINKED ISSUE #177: Build issues
A couple of things ...

1. I think the gh-pages build should be triggered on every merge to master, even if the build fails. What do other think?
2. The most recent build failed (https://ci.appveyor.com/project/ColinEberhardt/containerjs/build/1.0.194), but the commit is showing a green tick https://github.com/symphonyoss/ContainerJS/commits/master

jleft:
- Simplify npm scripts
- Remove `test` tasks that, inconsistently, either printed a message or ran the build (rather than actually running any tests)
- Ensure all `build` tasks are run on CI (this wasn't happening for `api-bundle`)

This also fixes #177, so feel free to merge this instead of #190.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,5 @@ node_js:
   - "4"
   - "6"
   - "7"
+script:
+  - npm run ci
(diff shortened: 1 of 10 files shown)


BenLambertNcl (approved):
LGTM. Will close #190 and merge this instead.

PR #194, opened 2017-06-11
Update Electron build to include CommonJS bundle and remove main process entry point

jleft:


BenLambertNcl (approved):

packages/api-electron/package.json
```diff
@@ -2,14 +2,15 @@
   "name": "containerjs-api-electron",
   "version": "0.0.2",
   "description": "",
-  "main": "index.js",
+  "main": "build/lib/preload.js",
```
ColinEberhardt, line 5: Hmm ... I'm not so sure this is correct. This package is being executed from the electron command line app, which uses the `main` property here to determine the application entry point:

https://electron.atom.io/docs/tutorial/quick-start/#write-your-first-electron-app

packages/api-electron/package.json
```diff
@@ -2,14 +2,15 @@
...
   "module": "build/es/preload.js",
   "jsnext:main": "build/es/preload.js",
   "browser": "build/dist/containerjs-api.js",
   "scripts": {
     "clean": "rimraf build",
-    "build": "npm run clean && npm run build:es && rollup -c",
-    "build:es": "rimraf build/es && tsc",
+    "build": "npm run clean && npm run build:cjs && npm run build:es && rollup -c",
+    "build:es": "rimraf build/es && tsc -p tsconfig.json -m ES2015 --outdir ./build/es",
```
ColinEberhardt, line 12: Considering the comment above, in what context would people use these various builds?

packages/api-electron/package.json
jleft: Good point, and I haven't tested it, but I think that the path to the entry point of the application is [explicitly provided](https://github.com/symphonyoss/ContainerJS/blob/3d3a1c7cb9f130d7d6c8b986764e828ace9c9454/packages/api-electron/bin/ssf-electron#L8) in the CLI, so the `main` entry point will not be used:

```js
var mainFile = path.join(__dirname, '../main.js');

// add the location of the main JS file to the command line args
var args = [mainFile, '--enable-sandbox'].concat(process.argv);

var child = proc.spawn(electron, args, {stdio: 'inherit'})
```

packages/api-electron/package.json
jleft: This change means that the user would no longer have access to Electron's main process when using this package; I think it makes sense to do this, as there's no equivalent in OpenFin/Browser.

 I've detailed the use of each entry point/build here: https://github.com/symphonyoss/ContainerJS/issues/132#issuecomment-302962997 - if this needs clarifying or changing, please let me know :)

BenLambertNcl:
@ColinEberhardt @jleft This PR has been open for some time, is it now redundant?

ColinEberhardt:
Personally I think it is!

BenLambertNcl:
@ColinEberhardt Ok ill close it. @jleft feel free to reopen this if you feel we need to add this.

PR #195, opened 2017-06-12
WIP: Change Electron to use window.open

BenLambertNcl:
This is a work in progress for changing the Electron API to use window.open. This is so child windows are created in the same process as their parent, therefore allowing access to each others JavaScript context.

WIP because features string doesn't work yet, meaning tests also fail.

### DIFF
--- a/packages/api-browser/src/window.ts
+++ b/packages/api-browser/src/window.ts
@@ -23,6 +23,7 @@ class Window {
   innerWindow: any;
   eventListeners: any;
   id: string;
+  globalWindow: any;
 
   constructor(options, callback, errorCallback) {
     this.children = [];
@@ -31,12 +32,14 @@ class Window {
 
     if (!options) {
       this.innerWindow = window;
+      this.globalWindow = this.innerWindow;
       this.id = window.name;
       if (callback) {
         callback();
       }
     } else {
       this.innerWindow = window.open(options.url, options.name, objectToFeaturesString(options));
+      this.globalWindow = this.innerWindow;
       this.id = this.innerWindow.name;
       this.innerWindow.onclose = () => {
         removeAccessibleWindow(this.innerWindow.name);
--- a/packages/api-demo/package.json
+++ b/packages/api-demo/package.json
@@ -24,8 +24,8 @@
   },
   "homepage": "https://github.com/symphonyoss/ContainerJS#readme",
   "dependencies": {
-    "containerjs-api-bundle": "^0.0.1",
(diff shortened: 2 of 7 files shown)


ColinEberhardt:
@BenLambertNcl I guess this means that it is looking like a viable option to have ContainerJS adopt a more OpenFin-like model instead?

ColinEberhardt:
One thing to be wary of is that the electron behaviour you get from window.open is different when the sandbox is enabled:

https://github.com/electron/electron/blob/master/docs/api/sandbox-option.md

I must admit, I hadn't turned it on in my own investigations!

BenLambertNcl:
@ColinEberhardt Yeah we can use the OpenFin process model with this change. I think its much nicer this way, as we give the user a choice, rather than forcing them into 1 specific process model. As for the link I hadn't seen that! It seemed to have all of the methods it mentioned in the documentation on the object when I tested it. Will double check though.

BenLambertNcl:
@ColinEberhardt From further research, it seems that we might need to give up some functionality using `window.open` due to the lack of window options that are supported when creating a window. The list that I have made of options that do not work with `window.open` are:
* `maxWidth`
* `maxHeight`
* `minWidth`
* `minHeight`
* `alwaysOnTop`
* `backgroundColor`
* `center`
* `frame`
* `maximizable`
* `minimizable`
* `resizable`
* `show`
* `skipTaskbar`
* `transparent`

Some of these can be set after the window has opened (such as resizable, minimizable etc.) but some like show and frame, only work when passed into the window constructor. It seems like some of these should work though, with references to show [here](https://github.com/electron/electron/blob/787bc8570382e98c4f204abff05b2af122e5a422/spec/fixtures/pages/window-open-hide.html#L4) as well as references to max/min width and others [here](https://github.com/electron/electron/blob/8ae36c250c456b7a334c00109ec9f5f01beab0a4/lib/browser/guest-window-manager.js#L193), but none do, even in [the electron api demos](https://github.com/electron/electron-api-demos). 

Not sure if it is worth continuing with this? Seems odd that these options are used in the source code and they do not work, could be a bug in Electron?

ColinEberhardt:
I must admit, I'm not too surprised - the MDN page for window.open does discourage its usage.

It might be worth parking this.

My preference is that you try to create a consistent process model between containers, but for the initial implementation I'm not too fussed about what that actually is.

Take the easiest path!

BenLambertNcl:
Ok I will stick to the "Each window as a separate process" like we were doing initially then. Ill close this PR for now, but we can always reopen it if we see value in using this later.

PR #199, opened 2017-06-13
Add symphony compatibility project

BenLambertNcl:
Adds a compatibility layer for mapping the [symphony API specifications](https://symphonyoss.atlassian.net/wiki/display/WGDWAPI/Proposed+Standard+API+Specifications) to ContainerJS APIs. most of the APIs are not implemented in ContainerJS yet, but it gives us a simple way of mapping them once they are added.

To use the compatibility layer, add the `containerjs-compatibility-api.js` file in a script tag in any webpage needing the compatibility. This could be improved by adding within the bundle/preload scripts.


packages/api-symphony-compatibility/src/mapping.ts
```diff
@@ -0,0 +1,56 @@
+const containerjsSsf = ssf;
+
+namespace map {
+  export class ssf {
+    static activate(windowName: string) {
+      // Could use window.focus in containerjs, but no way of selecting window by name yet
+      return;
+    }
+
+    static registerActivityDetection(throttle: number, callback: Function) {
```
ColinEberhardt: For these methods I think it would be useful to reference the symphony confluence URL where they are specifed

packages/api-symphony-compatibility/src/mapping.ts
```diff
@@ -0,0 +1,56 @@
+const containerjsSsf = ssf;
+
+namespace map {
+  export class ssf {
+    static activate(windowName: string) {
+      // Could use window.focus in containerjs, but no way of selecting window by name yet
```
ColinEberhardt: Can you iterate over all windows to find the one with the given name?

packages/api-demo/package.json
```diff
@@ -25,6 +25,7 @@
   "dependencies": {
     "containerjs-api-bundle": "0.0.2",
     "containerjs-api-electron": "0.0.2",
+    "containerjs-compatibility-api": "0.0.1",
```
ColinEberhardt: rather than adding to the existing api-demo, I'd prefer to keep this code self-container within `containerjs-compatibility-api` as I don't think we see these APIs being included in the long-term. Having it all in one place makes it easier to remove ;-)

ColinEberhardt (changes requested):
Looks really good in general - just a few minor changes. Thanks!

packages/api-symphony-compatibility/src/mapping.ts
BenLambertNcl: Not using ContainerJS. The only method we have is `getCurrentWindow`, but it should be relatively easy to add this in.

packages/api-demo/package.json
BenLambertNcl: I only included this as a simple demo for how it works. Do you want to remove the demo, or add a new demo in the `containerjs-compatibility-api` project?

packages/api-symphony-compatibility/src/mapping.ts
ColinEberhardt: OK, fair enough - that does look like a bit of an omission in our current API - and is something covered in #171

packages/api-demo/package.json
ColinEberhardt: I'd prefer to have a new demo in containerjs-compatibility-api - thanks

BenLambertNcl:
@ColinEberhardt Updated :-)

ColinEberhardt (approved):

PR #205, opened 2017-06-15
Fix small errors in documentation pages

### LINKED ISSUE #204: Documentation has wrong return description
image

The return description in the documentation is the same as the method description, when it should be the return description. (in this case: "A promise which resolves to nothing when the function has completed.")

BenLambertNcl:
Closes #204


packages/api-specification/transform-type-info.js
```diff
@@ -150,7 +150,7 @@ const documentClass = (className) => {
           h5('Returns'),
           dl([
             dt(formatType(d.match.type), {class: 'code return-value'}),
-            dd(formatComment(d.match.comment))
+            dd(d.match.comment ? d.match.comment.returns : '')
```
ColinEberhardt, line 153: I wasn't sure what the 'schema' for comments was, which is why I pushed everything through the `formatComment` function.

Would it work if you just fixed the mistake I made in the function ...

comment.return => comment.returns

ColinEberhardt (approved):
One small comment, but otherwise looks good,

packages/api-specification/transform-type-info.js
BenLambertNcl: Yeah that worked, but for some methods it would print both the method description and the return text, so the method text was still being repeated. e.g.

```
"comment": {
  "shortText": "Flashes the window's frame and taskbar icon.",
  "returns": "A promise which resolves to nothing when the function has completed.\n"
}
```
Displays both text elements under the return type like so:

---

### flashFrame
Flashes the window's frame and taskbar icon.

**Returns:** `Promise<void>`
Flashes the window's frame and taskbar icon.
A promise which resolves to nothing when the function has completed.

---
Which looked a bit odd to me.

PR #207, opened 2017-06-20
Add hidden window to OpenFin API to track window states

### LINKED ISSUE #123: Update OpenFin to use new preload script
In the newest [OpenFin Update](https://developer.openfin.co/versions/?product=Runtime%20Versions&version=6.49.20.21), a preload script option was added. It would probably be worth using this functionality for our OpenFin demo to make it more consistent with Electron.

### LINKED ISSUE #134: getCurrentWindow potentially loses children windows
The current implementation of `getCurrentWindow` creates a new `ssf.window` wrapper around the `innerWindow`. This means that it is possible (and likely) to have two `ssf.windows` pointing to the same `innerWindow` however they will have different children and listeners to each other. 

The parent that creates a window and the child window itself both have their own wrapper around the `innerWindow`. So once the child creates another window or adds a listeners, the parent still has a wrapper that says there are no children or listeners.

I have investigated a few solutions however none of them seem ideal. One would be to put a reference of the child on the parent window and have the child use `window.opener` to get the currenty `ssf.window` rather than creating a new one, however this may not work at all for electron.

Another solution could be to have a higher up window or manager window that isnt shown and is a state of truth for all the other windows so that their references don't get out of sync. This solution could work however it would be a big task.

BenLambertNcl:
Adds a hidden window in OpenFin to keep track of the window state, communicating via IPC. Because of this, the `getParentWindow()` method has been made async. The new preload script has also been added, as any windows that were created and did not include the containerjs-api (i.e. github.com) would be lost by the hidden window.

Also, the `getParentWindow` test has been skipped for the time being, as in OpenFin, doing `selectWindow(2)` selects the new hidden window, which means ssf is not defined and the test fails. This could be fixed once the startup for the containers has changed, so we can feed OpenFin the app.json file for a hidden window first, which can then create a new window with the config that was passed in.

Closes #134 and closes #123.

### DIFF
--- a/packages/api-demo/src/app.json
+++ b/packages/api-demo/src/app.json
@@ -4,7 +4,8 @@
         "name": "SSF Desktop Wrapper",
         "url": "http://localhost:8080/index.html",
         "uuid": "ssf-desktop-api-openfin-demo",
-        "autoShow": true
+        "autoShow": true,
+        "preload": "http://localhost:8080/scripts/containerjs-api-bundle.js"
     },
     "runtime": {
         "version": "stable"
--- a/packages/api-openfin/src/main-process.ts
+++ b/packages/api-openfin/src/main-process.ts
@@ -1,0 +1,127 @@
+import MessageService from './message-service';
+
+// Code that is "evaled" when the main window has been opened, sets up
+// all the InterApplicationBus listeners for window events to keep track of state
+const mainWindowCode = () => {
+  const childTree = [];
+
+  const addParent = (parentName, windowName, tree) => {
+    let win = tree.find(w => w.name === parentName);
+    if (win) {
+      win.children.push({
+        name: windowName,
+        children: []
+      });
+      return true;
+    } else {
+      return tree.some((childWin) => {
+        return addParent(parentName, windowName, childWin.children);
+      });
+    }
+  };
+
+  const getParentWindow = (name, parentName, tree) => {
+    let win = tree.find(w => w.name === name);
+    if (win) {
+      return parentName;
+    } else {
+      let parentName = null;
+      tree.some((childWin) => {
+        parentName = getParentWindow(name, childWin.name, childWin.children);
+        return parentName !== null;
+      });
+      return parentName;
+    }
+  };
+
+  const getChildWindows = (name, tree) => {
+    let win = tree.find(w => w.name === name);
+    if (win) {
+      return win.children.map(c => c.name);
+    } else {
+      let childNames = [];
+      tree.some((childWin) => {
+        childNames = getChildWindows(name, childWin.children);
+        return childNames && childNames.length > 0;
+      });
+      return childNames;
+    }
+  };
+
+  const deleteWindow = (name, tree) => {
+    let index = tree.findIndex(w => w.name === name);
+    if (index >= 0) {
+      tree.splice(index, 1);
+    } else {
+      tree.forEach((childWin) => {
+        deleteWindow(name, childWin.children);
+      });
+    }
+  };
+
+  fin.desktop.InterApplicationBus.subscribe('*', 'ssf-new-window', (data) => {
+    if (data.parentName === null) {
+      childTree.push({
+        name: data.windowName,
+        children: []
+      });
+    } else {
+       if (!addParent(data.parentName, data.windowName, childTree)) {
+        // No parent in tree, make a new one
+        childTree.push({
+          name: data.parentName,
+          children: [{
+            name: data.windowName,
+            children: []
+          }]
+        });
+      }
+    }
+  });
+
+  fin.desktop.InterApplicationBus.subscribe('*', 'ssf-get-parent-window', (name, uuid) => {
+    const parent = getParentWindow(name, null, childTree);
+    fin.desktop.InterApplicationBus.send(uuid, 'ssf-parent-window', parent);
+  });
+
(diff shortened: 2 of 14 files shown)


packages/api-demo/src/app.json
```diff
@@ -4,7 +4,8 @@
         "name": "SSF Desktop Wrapper",
         "url": "http://localhost:8080/index.html",
         "uuid": "ssf-desktop-api-openfin-demo",
-        "autoShow": true
+        "autoShow": true,
+        "preload": "http://localhost:8080/scripts/containerjs-api-bundle.js"
```
ColinEberhardt, line 8: Can you also update the documentation here:

https://github.com/symphonyoss/ContainerJS/blob/master/docs/index.md

packages/api-openfin/src/main-process.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  const closeChildren = (uuid) => {
+    const childUuids = getChildWindows(uuid, childTree);
+    childUuids.forEach((child) => {
+      fin.desktop.Application.wrap(child).close(true);
+      closeChildren(child);
+    });
+  }
+
+  fin.desktop.InterApplicationBus.subscribe('*', 'ssf-new-window', (data) => {
```
ColinEberhardt, line 70: As this bus can be used by multiple applications, I think that rather than using `'*'` you should be using some sort of UUID that is specific to this (containerjs) application.

packages/api-openfin/src/window.ts
```diff
@@ -236,18 +245,20 @@ class Window implements ssf.Window {
...
   getParentWindow() {
     return new Promise((resolve, reject) => {
-      if (this.innerWindow) {
-        let parent = null;
-        if (window.opener) {
-          const win = fin.desktop.Window.wrap(this.innerWindow.uuid, window.opener.name);
-          parent = new Window(null, null, null);
-          parent.innerWindow = win;
+      fin.desktop.InterApplicationBus.publish('ssf-get-parent-window', this.innerWindow.uuid);
+      fin.desktop.InterApplicationBus.subscribe('*' , 'ssf-parent-window', (name) => {
```
ColinEberhardt: What happens if `getParentWindow` is invoked more than once? do you get duplicate subscriptions?

packages/api-demo/src/app.json
ColinEberhardt, line 8: On another note, is there any reason why the api-bundle is being used as the OpenFin preload script? I'd have thought you'd just want to preload the OpenFin specific code?

(and if OpenFin and Electron both use preload scripts, does that mean the api-bundle is now redundant?)

ColinEberhardt, line 8: And another question! Does this `preload` property ensure that this scripts is preloaded for any window opened via `new fin.desktop.Window()`?

ColinEberhardt (commented):
Nice job - a few questions, but looks good to me.

packages/api-demo/src/app.json
BenLambertNcl: Yeah, windows opened via `window.open` and `fin.desktop.Window` inherit the preload of the parent application. For applications this isn't true, but we can pass the current applications preload script in (i.e. `fin.desktop.Window.gerCurrent().getOptions(o => preloadScriptUrl = o.preload)`) which is what i'm doing currently

packages/api-openfin/src/main-process.ts
BenLambertNcl: The problem doing this is that multiple windows need to be able to send to this window, and because each window is now a new application they all have a different UUID. Especially as before this point, we don't know a new window exists (as this is telling the `mainWindow` of a new window) so having a list of UUIDs in our application wouldn't help. Unless you can think of a different way around it?

packages/api-openfin/src/window.ts
BenLambertNcl: Ah yeah good spot!

packages/api-openfin/src/main-process.ts
ColinEberhardt: Ah, just checked the docs, I hadn't realised that this filter was for the application UUID, I thought it was just some sort of topic string.

Does that mean you will have to filter these messages to ensure you only respond to those relating to your current ContainerJS application?

BenLambertNcl: Im not sure whether its only applications in our ContainerJS application or if any applications running on the OpenFin RVM could send it messages (i.e. if multiple OpenFin apps were running at once). Ill look into it, as I can't seem to find it in the docs

ColinEberhardt: I have a feeling it would be any application running on the OpenFin RVM as implied by the InterApplicationBus name - I know that inter-app communication is one of the advertised features of OpenFin.

On this note, I'd be happy for you to leave the code as-is, but raise this as an issue for future resolution.

packages/api-demo/src/app.json
BenLambertNcl: Yeah possibly, I hadn't thought of that. It still works the same as before like this, but you're right in that the `bundle.js` wouldn't be in the html anymore so we wouldn't need to check for `fin` being defined or anything, making the bundle redundant.

BenLambertNcl:
Updated this PR. Ill add 2 new issues (one for the possible removal of the api-bundle project and 1 for possibly filtering window events for our application only)

ColinEberhardt (approved):
Thanks for the updates - I'd be happy to see this merged 👍

PR #214, opened 2017-06-21
Fix multiple inconsistencies between windows

### LINKED ISSUE #209: new ssf.Window on electron doesn't support relative URLs
If I try to open a window as follows:

```
const window = new ssf.Window({ url: 'popup.html' });
```

it fails (silently), although the following works:

```
const window = new ssf.Window({ url: 'http://localhost:8080/popup.html' });
```

I think relative URLs should be supported. They work just fine in the browser.

### LINKED ISSUE #210: new ssf.Window has different required options than the browser / electron
In order to get a new window to show with OpenFin, I need to add the following options:

```
const window = new ssf.Window({
  url: 'http://localhost:8080/popup.html',
  name: 'bob',
  autoShow: true
});
```

Whereas electron & the browser just require the `url`

### LINKED ISSUE #211: new ssf.Window has a different default location with OpenFin
With electron a new window defaults to having the same location / bounds as the parent:
```
const window = new ssf.Window({ url: 'http://localhost:8080/popup.html', name: 'bob', autoShow: true });
```

With OpenFin it does not.

BenLambertNcl:
- Default name for OpenFin windows is a random GUID
- Relative URLs are checked in both Electron and OpenFin

Closes #209, closes #210 and closes #211.

### DIFF
--- a/packages/api-electron/src/preload/window.ts
+++ b/packages/api-electron/src/preload/window.ts
@@ -8,12 +8,26 @@ import MessageService from './message-service';
 import { IpcMessages } from '../common/constants';
 
 let currentWindow = null;
+const isUrlPattern = /^https?:\/\//i;
 
 class Window implements ssf.Window {
   innerWindow: Electron.BrowserWindow;
   id: string;
 
   constructor(options: ssf.WindowOptions, callback, errorCallback) {
+    if (!isUrlPattern.test(options.url) && options.url !== 'about:blank') {
+      if (options.url.startsWith('/')) {
+        // File at root
+        options.url = location.origin + options.url;
+      } else {
+        // relative to current file
+        const pathSections = location.pathname.split('/').filter(x => x);
+        pathSections.splice(-1);
+        const currentPath = pathSections.join('/');
+        options.url = location.origin + '/' + currentPath + options.url;
+      }
+    }
+
     MessageService.subscribe('*', 'ssf-window-message', (...args) => {
       const event = 'message';
       this.innerWindow.emit(event, ...args);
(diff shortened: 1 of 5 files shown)


packages/api-electron/src/preload/window.ts
```diff
@@ -28,6 +29,19 @@ class Window implements ssf.Window {
       return this;
     }
 
+    if (!isUrlPattern.test(options.url) && options.url !== 'about:blank') {
```
ColinEberhardt: Can you add a comment above this block of code to explain why you are manipulating the URL?

Also, the code mutates the `option` object that the user passes in (which is a bit nasty), I think it would be better to move this code into a separate function that returns a new URL string.

packages/api-openfin/src/window.ts
```diff
@@ -38,10 +38,26 @@ const windowStates = {
...
+    return Math.floor((1 + Math.random()) * 0x10000)
+      .toString(16)
+      .substring(1);
+  }
+  return s4() + s4() + '-' + s4() + '-' + s4() + '-' +
+    s4() + '-' + s4() + s4() + s4();
+};
+
+const defaultOptions = {
+  name: guid(),
```
ColinEberhardt: As the `guid` function is invoked when the `defaultOptions` object is being constructed, doesn't this mean that all windows will be given the same name?

ColinEberhardt (approved):
Nice job :-)

A few minor comments, but otherwise, looks good to me.

packages/api-openfin/src/window.ts
BenLambertNcl: Yeah, good find!

packages/api-electron/src/preload/window.ts
BenLambertNcl: Oops, was meant to change that, thanks!

BenLambertNcl:
Updated :-)

ColinEberhardt (approved):
thanks for the updates - LGTM 👍

PR #215, opened 2017-06-22
Add git attributes file for ssf-electron binary

### LINKED ISSUE #196: ssf-electron Mac issues
The globally installed `ssf-electron` fails on Mac

```
$ npm install --global containerjs-api-electron
...
$ ssf-electron 
env: node\r: No such file or directory
```

This looks like a [line endings issue](https://stackoverflow.com/questions/30344858/node-script-executable-not-working-on-mac-env-node-r-no-such-file-or-directo).

When installing globally I can see CRLF line endings:

```
$ file .../containerjs-api-electron/bin/ssf-electron 
.../containerjs-api-electron/bin/ssf-electron: a /usr/bin/env node script text executable, ASCII text, with CRLF line terminators
```

And changing this to LF fixes the issue.

Although, my git clone of the project is showing LF endings - I guess the auto line ending setting of git is automagically changing line endings when I checkout.

BenLambertNcl:
Closes #196. It might be a good idea @ColinEberhardt to pull down this branch to make sure it fixes the problem. It seems to work fine on my machine with a fresh clone of the repo though :-)

### DIFF
--- a/packages/api-electron/.gitattributes
+++ b/packages/api-electron/.gitattributes
@@ -1,1 +1,2 @@
-bin/ssf-electron text eol=crlf
+# Needed for running the executable on mac/linux
+bin/ssf-electron text eol=lf


ColinEberhardt (changes requested):
You've got the `.gitattributes` file in the wrong folder ;-)

Also - can you add a comment that briefly explains why LF endings are required for this file?

BenLambertNcl:
@ColinEberhardt Oops! Thanks for that. Updated.

ColinEberhardt (approved):

PR #222, opened 2017-06-30
Make windows implement most event emitter methods

BenLambertNcl:
This adds event emitter methods (i.e. `on`, `once` etc.) to the window object. There are some methods of the `Node.EventEmitter` that could not be implemented (such as `prependListener` in OpenFin).

_Note:_ The Electron `BrowserWindow` object is an event emitter, but we couldn't use most of the methods it provided (such as `win.listeners(event)`) because the remote module adds listeners for multiple events to control the window from the main process. This would be inconsistent with OpenFin/Browser so the same implementation was used in Electron as in those APIs

### DIFF
--- a/packages/api-browser/src/window.ts
+++ b/packages/api-browser/src/window.ts
@@ -164,7 +164,7 @@ class Window implements ssf.WindowCore {
 
   listenerCount(event) {
     return this.eventListeners.has(event) ? this.eventListeners.get(event).length : 0;
-  }n
+  }
 
   listeners(event) {
     return this.eventListeners.get(event);
@@ -177,14 +177,7 @@ class Window implements ssf.WindowCore {
       listener();
     };
 
-    if (this.eventListeners.has(event)) {
-      const temp = this.eventListeners.get(event);
-      temp.push(listener);
-      this.eventListeners.set(event, temp);
-    } else {
-      this.eventListeners.set(event, [unsubscribeListener]);
-    }
-    this.innerWindow.addEventListener(eventMap[event], unsubscribeListener);
+    this.on(event, unsubscribeListener);
     return this;
   }
 
--- a/packages/api-browser/src/window.ts
+++ b/packages/api-browser/src/window.ts
@@ -14,7 +14,7 @@ const getWindowOffsets = (win) => {
 class Window implements ssf.WindowCore {
   children: ssf.Window[];
   innerWindow: any;
-  eventListeners: any;
+  eventListeners: Map<string, ((...args: any[]) => void)[]> = new Map();
   id: string;
 
   constructor(options, callback?, errorCallback?) {
--- a/packages/api-browser/src/window.ts
+++ b/packages/api-browser/src/window.ts
@@ -151,6 +151,41 @@ class Window implements ssf.WindowCore {
       this.eventListeners.set(event, [listener]);
     }
     this.innerWindow.addEventListener(eventMap[event], listener);
+    return this;
+  }
+
+  on(event, listener) {
+    return this.addListener(event, listener);
+  }
+
+  eventNames() {
+    return Array.from<string>(this.eventListeners.keys());
+  }
+
+  listenerCount(event) {
+    return this.eventListeners.has(event) ? this.eventListeners.get(event).length : 0;
+  }n
+
+  listeners(event) {
+    return this.eventListeners.get(event);
+  }
+
+  once(event, listener) {
+    // Remove the listener once it is called
+    const unsubscribeListener = () => {
+      this.removeListener(event, unsubscribeListener);
+      listener();
+    };
+
+    if (this.eventListeners.has(event)) {
+      const temp = this.eventListeners.get(event);
+      temp.push(listener);
+      this.eventListeners.set(event, temp);
+    } else {
+      this.eventListeners.set(event, [unsubscribeListener]);
+    }
+    this.innerWindow.addEventListener(eventMap[event], unsubscribeListener);
+    return this;
   }
 
   removeListener(event, listener) {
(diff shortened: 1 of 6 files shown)


packages/api-browser/src/window.ts
```diff
@@ -151,6 +151,34 @@ class Window implements ssf.WindowCore {
...
+
+  listeners(event) {
+    return this.eventListeners.get(event);
+  }
+
+  once(event, listener) {
+    // Remove the listener once it is called
+    const unsubscribeListener = () => {
+      this.removeListener(event, unsubscribeListener);
+      listener();
```
andy-lee-eng: Should this pass through the `Event` parameter?
i.e.
````
const unsubscribeListener = (evt) => {
  this.removeListener(event, unsubscribeListener);
  listener(evt);
}
````

File: packages/api-electron/src/preload/window.ts [hunk 2]
```diff
@@ -242,15 +243,77 @@ class Window implements ssf.Window {
...
+    return this;
   }
 
-  removeAllListeners() {
-    this.innerWindow.removeAllListeners();
+  removeAllListeners(eventName) {
+    const removeAllListenersForEvent = (event) => {
+      if (this.eventListeners.has(event)) {
+        this.eventListeners.get(event).forEach((listener) => {
+          this.innerWindow.removeListener(event, listener);
```
andy-lee-eng, line 304: There's a build error here, because `this.eventListeners.get(event)` is defined as a collection of `Function`, whereas `this.innerWindow.removeListener(event, listener)` expects `listener` to be of type `(...args: any[]) => void` (from `Electron.BrowserWindow`).

The problem doesn't arise in the above `removeListener` function because `listener` is type `any` (and TypeScript doesn't seem to mind that it doesn't match the interface definition, which specifies `Function`).

i.e. If you change `removeListener` to match the interface like this...
```
removeListener(event: string, listener: Function) {
...
```
... you'll get the same error there.

andy-lee-eng: In the openfin implementation, the problem doesn't arise because `this.eventListeners` is defined as just `Map<any, any>`, rather than the more strongly typed `Map<string, Function[]>`

packages/api-specification/interface.ts
```diff
@@ -117,6 +117,59 @@ declare namespace ssf {
     transparent?: boolean;
   }
 
+  class EventEmitter {
+    /**
+     * Adds a lister that runs when the specified event occurs. Alias for <span class="code">on()</span>.
```
andy-lee-eng: Should this be "listener" rather than "lister"? (same typo in the other function docs too)

packages/api-tests/test/window.core.spec.js
```diff
@@ -331,6 +358,74 @@ describe('WindowCore API', function(done) {
...
+        () => selectWindow(app.client, 1),
+        () => callAsyncWindowMethod('blur'),
+        () => callAsyncWindowMethod('focus'),
+        () => callWindowMethod('listenerCount', 'focus'),
+        (result) => assert.equal(result.value, 0)
+      ];
+
+      return chainPromises(steps);
+    });
+
```
andy-lee-eng, line 428: Is it worth including tests for removeListener and removeAllListeners (the latter with and without an event name)?

andy-lee-eng (changes requested):
Some TypeScript build errors that need looking into.

packages/api-electron/src/preload/window.ts
BenLambertNcl: Thanks! I think changing all the APIs to use the more specific type is better, so the full type will be `Map<string, ((...args: any[]) => void)[]>`

BenLambertNcl:
@DevAndyLee Thanks for the review. Updated the PR.

packages/api-openfin/src/window.ts
```diff
@@ -129,7 +129,7 @@ const convertOptions = (options: ssf.WindowOptions) => {
 };
 
 class Window implements ssf.Window {
-  eventListeners: Map<any, any>;
+  eventListeners: Map<string, ((...args: any[]) => void)[]> = new Map();
```
andy-lee-eng, line 132: We also do `this.eventListeners = new Map();` in the constructor, which is now redundant.

packages/api-tests/test/window.core.spec.js
```diff
@@ -331,6 +358,200 @@ describe('WindowCore API', function(done) {
...
+        /* eslint-enable no-undef */
+        return executeAsyncJavascript(app.client, script);
+      };
+
+      const steps = [
+        ...setupWindowSteps(windowOptions),
+        () => addWindowListener('blur', 'one'),
+        () => addWindowListener('blur', 'two'),
+        () => addWindowListener('focus', 'one'),
+        () => removeWindowListeners('blur'),
```
andy-lee-eng: Should be `removeWindowListeners()` (the function doesn't take a parameter).

packages/api-tests/test/window.core.spec.js
```diff
@@ -331,6 +358,200 @@ describe('WindowCore API', function(done) {
...
+        };
+        /* eslint-enable no-undef */
+        return executeAsyncJavascript(app.client, script, event);
+      };
+
+      const steps = [
+        ...setupWindowSteps(windowOptions),
+        () => addWindowListener('blur', 'one'),
+        () => addWindowListener('blur', 'two'),
+        () => removeWindowListeners('blur'),
```
andy-lee-eng: Maybe we could have `() => addWindowListener('focus', 'one')` here, and verify that the listener for `focus` **doesn't** get removed?

andy-lee-eng (approved):
A couple of comments that I think would improve things a bit, but otherwise this is looking good.

packages/api-tests/test/window.core.spec.js
BenLambertNcl: Yeah good idea :+1:

BenLambertNcl:
@DevAndyLee Updated and rebased. Happy to merge? :-)

andy-lee-eng:
yep :-)

PR #225, opened 2017-07-04
Fix issue with opening the child window in the messaging demo

andy-lee-eng:
The html file is now in a sub folder. The new code also doesn't assume
that the application is running on the localhost.
The label change makes it consistent with the parent window.


BenLambertNcl (approved):
Thanks :+1:

andy-lee-eng:
Apologies - I just realised I was logged into git with the wrong user id. Hopefully fixed now and committed as the correct user.

BenLambertNcl:
LGTM :+1:, I assume you have permissions to merge once you rebase/update the branch?

PR #227, opened 2017-07-04
Quick-fix to allow the demo app to access notification.html

andy-lee-eng:
OpenFin notifications require access to this file.

### DIFF
--- a/packages/api-demo/package.json
+++ b/packages/api-demo/package.json
@@ -5,7 +5,7 @@
   "main": "src/index.js",
   "scripts": {
     "serve:no-browser": "npm run serve -- --no-browser",
-    "serve": "live-server --mount=/:src --mount=/scripts/containerjs-api-bundle:node_modules/containerjs-api-bundle/build/containerjs-bundle.js",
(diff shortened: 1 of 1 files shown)


packages/api-demo/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "src/index.js",
   "scripts": {
     "serve:no-browser": "npm run serve -- --no-browser",
-    "serve": "live-server --mount=/:src --mount=/scripts/containerjs-api-bundle:node_modules/containerjs-api-bundle/build/containerjs-bundle.js",
+    "serve": "live-server --mount=/:src --mount=/scripts/containerjs-api-bundle:node_modules/containerjs-api-bundle/build/containerjs-bundle.js --mount=/notifications/notification.html:node_modules/containerjs-api-bundle/build/notification.html",
```
BenLambertNcl, line 8: Think we should be able to do `--mount=/scripts/containerjs-api-bundle:node_modules/containerjs-api-bundle/build/` to serve both files together

BenLambertNcl (approved):
Thanks! Think its possible to make the command a little nicer though :)

packages/api-demo/package.json
andy-lee-eng: I tried that first actually, and it didn't work.

The js file is served as the url `/scripts/containerjs-api-bundle` (without `.js`).
The template file is expected relative to the screen, so `/notifications/notification.html`

As discussed this morning, this is perhaps an area where the API needs looking at anyway, so I'd hope this would change to something better.

PR #228, opened 2017-07-05
Add cli wrapper around OpenFin

BenLambertNcl:
Adds a pretty basic wrapper around the OpenFin cli, the idea being that in the future, this would give us the flexibility to change the format of the `app.json` that we use.


andy-lee-eng (approved):

ColinEberhardt:
@DevAndyLee @BenLambertNcl one thing to consider with changes like this is to update the website / documentation too!

BenLambertNcl:
@ColinEberhardt Oops! Ill make a PR for the docs update ASAP.

andy-lee-eng:
Yes, good point. That did actually occur to me as I was looking at it but then I got distracted by something else and forgot to add a comment :-(

PR #229, opened 2017-07-05
Update documentation to reflect OpenFin cli change

BenLambertNcl:


docs/index.md
```diff
@@ -73,16 +68,16 @@ This starts a server on port 8080.
 
 #### Running with OpenFin
 
-OpenFin has a command line tool for launching OpenFin applications. This can be installed as follows:
+The ContainerJS project provides an OpenFin-based command line tool, which can be installed as follows:
```
andy-lee-eng, line 71: Is this actually true, or is this something we plan to do?

As far as I can tell the ssf-openfin command is available via the containerjs-api-openfin dependency (so you can run `ssf-openfin...` as a script in package.json), but you can't actually install it globally with that command. Same goes for ssf-electron actually.

BenLambertNcl: I assumed the electron worked so copied that, but you're right, that wouldn't work. Ill change the docs for both electron + openfin.

BenLambertNcl:
@DevAndyLee Updated. :)

andy-lee-eng (approved):

docs/index.md
ColinEberhardt: You can actually instal `ssf-electron` globally, it's also available via npm, so can be installed globally from there.

BenLambertNcl: @DevAndyLee I misread your comment yesterday. If we install the package globally (i.e. `npm i -g containerjs-api-openfin`) we can then use `ssf-openfin` globally. The documentation was still incorrect I think (It was saying to install `ssf-electron` directly from npm, when we need to install `containerjs-api-electron` instead). Is that correct @ColinEberhardt?

ColinEberhardt: Oh yes, you're absolutely correct, the `containerjs-api-electron` package exposes the `ssf-electron` command

PR #231, opened 2017-07-06
Implemented request permission behaviour for Electron and Browser

andy-lee-eng:
Gives a consistent API to call for all three containers, but Electron and
OpenFin automatically grant permission without asking the user.

Defined interface to match the existing standard API in the browser.


andy-lee-eng:
It means that now the client code can just do something simple like:
```javascript
Notification.requestPermission(permission => {
  if (permission === 'granted') {
    new Notification('my-title', { body: 'some-content' })
  }
})
```
and that will work in all containers.

It still uses a shim for Window.Notification though, rather than defining a new API like we do for things like ssf.Window etc...

packages/api-electron/src/preload/shims.ts
```diff
@@ -1,13 +1,23 @@
 const remote = require('electron').remote;
 const eNotify = remote.require('electron-notify');
+const PERMISSION_GRANTED = "granted";
```
ColinEberhardt: perhaps now would be a good time to rename this file from `shims.ts` to something that follows the naming convention for the other files in this folder?

packages/api-specification/interface.ts
```diff
@@ -583,6 +583,23 @@ declare namespace ssf {
...
+  class Notification {
+    /**
+     * Create a notification
+     * @param title - The title text of the notification.
+     * @param options - The notification options.
+     */
+    constructor(title: string, options: NotificationOptions);
+    
+    static permission: string;
+    static requestPermission(callback: (NotificationPermissionCallback));
```
ColinEberhardt: Looking on MDN the callback signature is deprecated. How about using the promise-based signature instead?

https://developer.mozilla.org/en-US/docs/Web/API/Notification/requestPermission

ColinEberhardt (approved):
Looks good to me - just a couple of comments

packages/api-electron/src/preload/shims.ts
andy-lee-eng: So `notification.ts` I guess.

Should we still be "shimming" the browser API like this then? For the other APIs we've exported a class under the `ssf` namespace.

i.e. instead of shimming `window.Notification`, should we just export `ssf.Notification` as a class, and implement that in all three containers?

ColinEberhardt:
> It still uses a shim for Window.Notification though, rather than defining a new API like we do for things like ssf.Window etc...

Yeah, that's an interesting observation. 

The API we are currently providing is the same as the HTML5 spec, however, we do want to add extra features. Perhaps using `ssf.Notification` would be a better approach?

Also, SymphonyElectron uses this namespace:

https://github.com/symphonyoss/SymphonyElectron/blob/299e75eca328375468cc3d0bf34ae9ca73e445f6/js/preload/preloadMain.js#L76

packages/api-specification/interface.ts
```diff
@@ -583,6 +583,23 @@ declare namespace ssf {
...
+
+  class Notification {
+    /**
+     * Create a notification
+     * @param title - The title text of the notification.
+     * @param options - The notification options.
+     */
+    constructor(title: string, options: NotificationOptions);
+    
+    static permission: string;
```
BenLambertNcl: I think we could add typedoc comments here so we get descriptions for the documentation. Might be good to have `static permission: 'granted' | 'denied' | 'default'` as well so we can see the possible values.

BenLambertNcl (commented):
LGTM! One small comment. I agree that we should move to `ssf.Notification` to keep in line with the other APIs.

BenLambertNcl (approved):

andy-lee-eng:
I've refactored to be ssf.Notification, and used the Promise based function.

I've also removed the `permission` getter, since it isn't part of the TypeScript definition on the Notification object in the es6 typescript file. For client code it's much cleaner to always call the promise function rather than check the property to decide whether to call the promise.

BenLambertNcl (approved):
LGTM!

ColinEberhardt:
:+1:

PR #234, opened 2017-07-06
Fix for console error showing in electron and browser

### LINKED ISSUE #233: Browser/Electron getting error "fin is not defined"
```
containerjs-api-bundle:172 Uncaught ReferenceError: fin is not defined
    at Function.MessageService.subscribe (containerjs-api-bundle:172)
    at new Window (containerjs-api-bundle:313)
    at Function.Window.getCurrentWindow (containerjs-api-bundle:648)
    at containerjs-api-bundle:654
    at childTree (containerjs-api-bundle:4)
    at containerjs-api-bundle:5
```

The problem seems to be some code at the end of /packages/api-openfin/src/window.js that always executes (and expects window.fin to exist)

BenLambertNcl:
I'm not sure if there is a better way to do this, would be nice if we could wrap all of the OpenFin API with `if ('fin' in window)` but not sure if roll-up supports this sort of thing. I think it makes sense when developing that the code written in `api-openfin` would only ever be executed in OpenFin.

Closes #233

### DIFF
--- a/packages/api-openfin/src/main-process.ts
+++ b/packages/api-openfin/src/main-process.ts
@@ -114,6 +114,13 @@ const mainWindowCode = () => {
 };
 
 const createMainProcess = (done) => {
+  // Populate the current window variable
+  ssf.Window.getCurrentWindow();
+
+  fin.desktop.InterApplicationBus.subscribe('*', 'ssf-close-all', () => {
+    fin.desktop.Window.getCurrent().close(true);
+  });
+
   // Create the main window, if the window already exists, the success callback isn't ran
   // and the already open window is returned
   const app = new fin.desktop.Application({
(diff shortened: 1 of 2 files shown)


andy-lee-eng (commented):
The code looks similar to some other things in `main-process.ts`, which gets called as part of `ssf.app.ready()`  (i.e. subscribing to `InterApplicationBus` events etc).

I thought it might make sense to move it there. Also, `main-process.ts` only gets called after `fin.desktop.main` has completed, whereas this version looks like it may risk that fin.desktop isn't ready...

BenLambertNcl:
Yeah that's a good point actually, I hadn't thought of that. Moving it into there is probably a good option for now, but I think it would still be useful to wrap the APIs in that if.

BenLambertNcl:
@DevAndyLee Made those changes. :-)

andy-lee-eng (approved):

PR #235, opened 2017-07-07
Refector typescript interface into separate files

BenLambertNcl:
Split the interface up into separate files to make it easier to work on different parts of the API. Also updated some of the types that were displaying as 'UNKNOWN' in the documentation. Example:
[`eventNames`](https://symphonyoss.github.io/ContainerJS/Docs#EventEmitter) to
image

### DIFF
--- a/packages/api-specification/interface/window-extension.ts
+++ b/packages/api-specification/interface/window-extension.ts
@@ -1,8 +1,0 @@
-declare interface Window {
-  /**
-   * Create a notification
-   * @param title - The title text of the notification.
-   * @param options - The notification options.
-   */
-  Notification(title: string, options: ssf.NotificationOptions): void;
-}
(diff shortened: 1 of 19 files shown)


packages/api-specification/interface/window-extension.ts
```diff
@@ -0,0 +1,8 @@
+declare interface Window {
```
andy-lee-eng: Do we still need this window-extension? I thought it related to the old way of opening a notification (and we now have ssf.Notification instead)... unless there's another need for it?

packages/api-specification/interface/window.ts
```diff
@@ -117,16 +100,19 @@ declare namespace ssf {
     transparent?: boolean;
   }
 
-  class EventEmitter {
+  /**
+   * Exposes methods that allow subscribing to particular events
+   */
+  abstract class EventEmitter {
```
andy-lee-eng: I think other objects are likely to need EventEmitter (e.g. Notification's onClick event), so should this be moved out too?

andy-lee-eng (approved):

BenLambertNcl:
@DevAndyLee Thanks, fixed those issues :-)

PR #237, opened 2017-07-07
Added icon and image support for Notification

andy-lee-eng:
I've tried to get the display as similar as possible between the three containers, but there are some differences - browser doesn't let you control the image size (and it's big), while OpenFin doesn't seem to have a way to alter the height.


andy-lee-eng:
Is there a handy central place to put utility functions (like `getAbsoluteUrl`)?

packages/api-electron/src/preload/notification.ts
```diff
@@ -2,18 +2,47 @@ const remote = require('electron').remote;
...
     eNotify.notify({
       title: title,
-      text: options.body
+      text: options.body,
+      image: this.getAbsoluteUrl(options.image)
     });
   }
 
+  getAbsoluteUrl(url) {
+    if (url && !url.startsWith('http:') && !url.startsWith('http:')) {
```
BenLambertNcl: Guessing one of these should be `!url.startsWith('https:')`?

BenLambertNcl (commented):
Would it be possible to get a screenshot of the notifications? :-)

BenLambertNcl:
As for `getAbsoluteUrl`, there isn't a good place at the moment. A common folder/package could be very useful though, maybe its worth making an issue for it?

andy-lee-eng:
notifications

andy-lee-eng:
Obviously there's still room for some style tweaking - background colour, font style/size etc.

BenLambertNcl:
I think that looks great, at least electron and openfin look very similar. Im guessing we won't have much control over what the browser does though :(

BenLambertNcl (approved):

PR #242, opened 2017-07-11
Add static window wrap method

BenLambertNcl:
Adds a wrap method onto the Window class, that allows wrapping a native container window with the ContainerJS window object. Mainly useful internally for methods such as `getParentWindow()`, where the native container method returns a native window that we need to convert to a ContainerJS window.

### DIFF
--- a/packages/api-tests/test/window.core.spec.js
+++ b/packages/api-tests/test/window.core.spec.js
@@ -616,6 +616,31 @@ describe('WindowCore API', function(done) {
       return chainPromises(steps);
     });
 
+    it.only('Should wrap a native window in a containerjs window object #ssf.Window.wrap', function() {
+      const windowTitle = 'windownamewrap';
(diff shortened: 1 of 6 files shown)


andy-lee-eng (approved):

packages/api-tests/test/window.core.spec.js
```diff
@@ -616,6 +616,31 @@ describe('WindowCore API', function(done) {
       return chainPromises(steps);
     });
 
+    it.only('Should wrap a native window in a containerjs window object #ssf.Window.wrap', function() {
```
andy-lee-eng, line 619: @BenLambertNcl Just noticed this `it.only` was left in, preventing other tests from running.

BenLambertNcl: Ok, ill remove it from #243.

PR #243, opened 2017-07-12
Add method to get window by id

BenLambertNcl:
Builds on the work that @kirilpopov did in #171 ([see method](https://github.com/symphonyoss/ContainerJS/pull/171/files#diff-9c41d64a6a03fd01f24706ed86e50b7eR19)). Currently just uses a static method on the window class, but we could move this into a separate class (like the initial PR) if necessary.

### DIFF
--- a/packages/api-openfin/src/window.ts
+++ b/packages/api-openfin/src/window.ts
@@ -538,26 +538,10 @@ class Window implements ssf.Window {
 
     let app = null;
     let existsPromise: Promise<void>;
-    if (id.match(new RegExp(idRegex))) {
-      existsPromise = appExists(id.split(':')[0]).then((exists) => {
-        if (!exists) {
-          app = null;
-        } else {
-          app = fin.desktop.Application.wrap(id.split(':')[0]);
-        }
-      });
-    } else {
-      // Assume just app id was passed in
-      existsPromise = appExists(id).then((exists) => {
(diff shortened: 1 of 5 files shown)


packages/api-openfin/src/window.ts
```diff
@@ -524,6 +520,52 @@ class Window implements ssf.Window {
...
+
+    const idRegex = '(' +    // Start group 1
+            '[\w\d-]+' +     // At least 1 word character, digit or dash
+            ')' +            // End group 1
+            ':' +            // colon
+            '\\1';            // Same string that was matched in group 1
+
+    let app = null;
+    let existsPromise: Promise<void>;
+    if (id.match(new RegExp(idRegex))) {
```
andy-lee-eng: I think this could be improved by having a branch to determine the app id, then then separately doing the appExists -> wrap code (i.e. not having to duplicate that code).

andy-lee-eng (approved):

packages/api-openfin/src/window.ts
BenLambertNcl: Yeah good idea. Thanks!

BenLambertNcl:
@DevAndyLee Refactored and remove the `.only` tag in the wrap test.

andy-lee-eng (approved):
👍

PR #244, opened 2017-07-12
Add tslint to the project

BenLambertNcl:
Adds tslint with a fairly lax config from [Microsoft](https://github.com/Microsoft/TypeScript/blob/master/tslint.json), as we only had eslint, so none of out ts files were being checked. We can add stricter rules as we see fit.


andy-lee-eng (approved):
👍

PR #245, opened 2017-07-12
Add get all method to window api

BenLambertNcl:
Implements `Window.getAll` which is similar to @kirilpopov suggestion [here](https://github.com/symphonyoss/ContainerJS/pull/171/files#diff-9c41d64a6a03fd01f24706ed86e50b7eR11).

### DIFF
--- a/packages/api-electron/src/preload/window.ts
+++ b/packages/api-electron/src/preload/window.ts
@@ -361,11 +361,7 @@ class Window implements ssf.Window {
   }
 
   static getAll() {
-    return new Promise<Window[]>(resolve => {
-      const windows = [];
-      BrowserWindow.getAllWindows().forEach((win) => windows.push(Window.wrap(win)));
-      resolve(windows);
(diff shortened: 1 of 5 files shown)


packages/api-electron/src/preload/window.ts
```diff
@@ -359,6 +359,14 @@ class Window implements ssf.Window {
       resolve(null);
     });
   }
+
+  static getAll() {
+    return new Promise<Window[]>(resolve => {
+      const windows = [];
```
andy-lee-eng: Can we just use `map`?

i.e.
```javascript
const windows = BrowserWindow.getAllWindows().map(Window.wrap);
```

andy-lee-eng (approved):

BenLambertNcl:
@DevAndyLee Refactored and rebased :-). Thanks

andy-lee-eng (approved):
👍

PR #246, opened 2017-07-12
 Refactored event emitter common code into utility

andy-lee-eng:
Refactored event emitter.

Also:
Use containerjs-bundle.js in the tests (instead of containerjs-api.js direct from openfin/browser projects), since it bundles up the api-utility library with it (and this is also the way that the code is used at runtime)


BenLambertNcl (approved):
LGTM!

andy-lee-eng:
Rebased to fix merge conflicts

BenLambertNcl (approved):

PR #247, opened 2017-07-12
Fix re-run problem with OpenFin window event tests

andy-lee-eng:
Force the new window to restore and move to override OpenFin's memory of where the window used to be


BenLambertNcl (approved):
:+1: Thanks

andy-lee-eng:
Rebased

BenLambertNcl (approved):

PR #248, opened 2017-07-12
Implemented 'click' event for notifications

andy-lee-eng:
Make Notification an EventEmitter and implement in all three containers

### DIFF
--- a/packages/api-browser/src/notification.ts
+++ b/packages/api-browser/src/notification.ts
@@ -1,6 +1,19 @@
-class BrowserNotification implements ssf.Notification {
+import { Emitter } from 'containerjs-api-utility';
+
+class BrowserNotification extends Emitter implements ssf.Notification {
+  innerNotification: Notification;
+
   constructor(title: string, options: ssf.NotificationOptions) {
-    new Notification(title, options);
+    super();
+    this.innerNotification = new Notification(title, options);
+  }
+
+  innerAddEventListener(eventName: string, handler: (...args: any[]) => void) {
+    this.innerNotification.addEventListener(eventName, handler);
+  }
+
+  innerRemoveEventListener(eventName: string, handler: (...args: any[]) => void) {
+    this.innerNotification.removeEventListener(eventName, handler);
   }
 
   static requestPermission(): Promise<ssf.NotificationPermission> {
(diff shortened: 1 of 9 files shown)


BenLambertNcl (approved):
LGTM. Got a bit confused with the empty `innerAddEventListener` functions but rereading the code it seems like the correct behaviour, maybe worth adding a quick comment such as 
> Not needed as listeners are added in the parent class

But might not be necessary

andy-lee-eng:
Yep. The innerAddEventListener stuff was needed because the Browser/OpenFin/Electron versions of Window all did it slightly differently. They all needed to do something though, so I made it abstract to force the derived class to implement it.

Now that it seems two of the Notification implementations don't actually need them, I'm thinking it might be simpler to make them non-abstract (with a no-op default implementation), so we can just leave them out if not required.

BenLambertNcl:
Yeah that sounds like a good idea. Im happy to leave that for a separate PR and just merge this as is

andy-lee-eng:
It'll be quick to do, so I may as well just do it now. Update coming soon...

BenLambertNcl (approved):

andy-lee-eng:
Almost there - looks like I broke a bunch of other Window tests by putting that set-size code in.

BenLambertNcl:
Yeah im looking into it now. Fixed that, but there seems to be another problem with openfin tests timing out

andy-lee-eng:
OpenFin tests are all passing for me now.

BenLambertNcl:
LGTM, I have a fix for the OpenFin tests timing out too, adding `containerjs-api-openfin` as a dependency to `api-tests`. You can include it in this PR or I can make a separate one?

andy-lee-eng:
Ah, ok.
I removed that dependency (and replaced it with the browser/openfin bundle `containerjs-api-bundle` instead), so that the tests would run in the same way that the real application runs, and also because the `container-api-utility` library is included in the bundle but not in the openfin version.

I guess I still had `container-api-openfin` in my `node_modules` though. Do you know why the the dependency is still needed? Does it need it in order to run `OpenFin` maybe?

BenLambertNcl:
Yeah, its needed for the ssf-openfin cli that we use to launch openfin. If ssf-openfin was installed globally (like on my machine) the tests will pass. Probably the same scenario if it was still inside node_modules.

PR #249, opened 2017-07-13
Better docs for Notification interface

andy-lee-eng:


BenLambertNcl (approved):
:+1:

ColinEberhardt (approved):

ColinEberhardt:
Looks good - I guess the only other thing I'd like to see here is a a few notes about the differences between browser, OpenFin and Electron.

PR #252, opened 2017-07-14
Add screen API to get all displays

BenLambertNcl:
Builds on @kirilpopov's PR [here](https://github.com/symphonyoss/ContainerJS/pull/171/files#diff-0507cae5e2b77c7fb5c16ed9a0a09cc8R7). Could not add tests as there is no way of getting monitor information in node, and using the browsers `screen.width` etc is unreliable with multi monitor setups (we have this same problem with the browser API itself)

### DIFF
--- a/packages/api-specification/interface/screen.ts
+++ b/packages/api-specification/interface/screen.ts
@@ -3,25 +3,34 @@ declare namespace ssf {
     static getDisplays(): Promise<Array<Display>>;
   }
 
+  /**
+   * Information about the users display. Note that the Browser API does not support multiple
+   * displays, so it assumes the display the browser is running in is the only display.
+   */
   export interface Display {
+    /**
+     * Unique Id of the display.
+     */
     id: string;
 
+    /**
+     * Current rotation of the display, can be 0, 90, 180, 270.
+     */
     rotation: number;
 
+    /**
+     * How much the display has been scaled.
+     */
     scaleFactor: number;
 
+    /**
+     * If the display is the primary display
+     */
     primary: boolean;
 
-    bounds: ssf.Bounds;
-  }
-
-  export interface Bounds {
-    left: number;
-
-    right: number;
-
-    top: number;
-
-    bottom: number;
+    /**
+     * Bounds of the display.
+     */
+    bounds: ssf.Rectangle;
   }
 }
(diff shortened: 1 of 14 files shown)


packages/api-specification/interface/screen.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * Information about the users display. Note that the Browser API does not support multiple
+   * displays, so it assumes the display the browser is running in is the only display.
+   */
+  export interface Display {
+    /**
+     * Unique Id of the display.
+     */
+    id: string;
```
ColinEberhardt, line 14: Just out of interest, where might I use this ID? might be worth mentioning in the comment for this property.

ColinEberhardt (approved):
looks good to me. thanks

andy-lee-eng (approved):

packages/api-specification/interface/screen.ts
BenLambertNcl: Honestly i'm not 100% sure, but its something that both Electron and OpenFin return. I think if we add more display methods it would make more sense (i.e. checking the display returned from `getPrimaryDisplay()` or `getFocusedDisplay()` and checking against the list of all displays).

andy-lee-eng: Maybe if you tried to open a new window on a particular display?

BenLambertNcl: That's not something that is supported naively, although we could add that into our API. To do that currently, you would have to position the window within the bounds of the 2nd window (i.e. `x: 1920, y:0`), so yeah maybe that is a good idea (seems more intuitive than the bounds checking).

ColinEberhardt: I think it's worth adding - there's an open issue on that topic for electron: https://github.com/electron/electron/issues/1836

BenLambertNcl: Ok. If everyone is happy I will merge his as is and add an issue to add this functionality (along with updating the comment once the functionality is there)?

ColinEberhardt: already done that #253 - and yes, happy to see this merged

PR #254, opened 2017-07-14
Add types to API typescript files

### LINKED ISSUE #115: Add more typescript types instead of using any
Currently `any` has been used a lot when adding typescript, however a lot of these  could potentially get changed into real types.

BenLambertNcl:
Adds more types to the API, as well as adding some extra tslint rules. The new rules are:
* arrow-return-shorthand: Prefer `() => a` over `() => { return a; }`
* comment check-uppercase: Start comments with uppcase letter
* eofline: Force new line at end of file
* no-default-export: disable `export default Window` in favor of `export class Window`, see [here](https://github.com/palantir/tslint/issues/1182#issue-151780453)
* only-arrow-functions: Require functions to be ES6 arrow functions not old style `function() {}`
* typedef: require type definitions for parameters, member variables and properties.

Closes #115.

### DIFF
--- a/packages/api-browser/index.ts
+++ b/packages/api-browser/index.ts
@@ -1,9 +1,9 @@
-import app from './src/app';
-import MessageService from './src/message-service';
-import ScreenSnippet from './src/screen-snippet';
-import Window from './src/window';
-import Notification from './src/notification';
-import Screen from './src/screen';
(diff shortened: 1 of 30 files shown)


andy-lee-eng (approved):
Fantastic 👍

PR #255, opened 2017-07-14
Added template option for Notification

andy-lee-eng:
Supports OpenFin's requirement for a template
Defaults to notification.html relative to current URL
Fixed bug in uri utility

### DIFF
--- a/packages/api-demo/package.json
+++ b/packages/api-demo/package.json
@@ -6,7 +6,7 @@
   "private": true,
   "scripts": {
     "serve:no-browser": "npm run serve -- --no-browser",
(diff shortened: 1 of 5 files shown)


BenLambertNcl (approved):
Looks good :-). Think this will be much nicer to use once we have a way of checking the container we are in (e.g. `app.container: 'openfin' | 'electron' | 'browser'`).

Ogadai:
Maybe, but as it's ignored in browser and electron (and assuming you want your app to work in all three), you might as well just set the setting no matter what container you're in.

I guess I could have called it `openfinTemplate`, but in case we decide to start using the template in Electron too...

andy-lee-eng:
Maybe, but as it's ignored in browser and electron (and assuming you want your app to work in all three), you might as well just set the setting no matter what container you're in.

I guess I could have called it openfinTemplate, but in case we decide to start using the template in Electron too...

PR #260, opened 2017-07-17
implemented setBadgeCount (for Mac)

ColinEberhardt:
Currently delegates straight through to Electron's app.setBadgeCount.

Interestingly SymphonyElectron provide a Windows implementation, looks like it is an overlay on top of the app icon:

https://github.com/symphonyoss/SymphonyElectron/blob/master/js/badgeCount.js


packages/api-demo/src/app/app-api-demo.js
```diff
@@ -0,0 +1,5 @@
+ssf.app.ready().then(() => {
+  document.getElementById('set-badge-count').onclick = () => {
+    ssf.app.setBadgeCount(Number(document.getElementById('badge-count').value));
+  };
+});
```
BenLambertNcl: New-line

packages/api-electron/src/preload/app.ts
```diff
@@ -1,5 +1,14 @@
+const {
+  remote
+} = require('electron');
+const electronApp = remote.app;
```
BenLambertNcl: If we only need one import here, might be better just to have:
`const electronApp = require('electron').remote.app`

BenLambertNcl (approved):
Very minor comments, looks good!

ColinEberhardt:
Thanks @BenLambertNcl - updated with your suggestions

BenLambertNcl (approved):

andy-lee-eng (approved):

PR #261, opened 2017-07-17
Fix and standardise emitter `emit` parameters

andy-lee-eng:
Change `postMessage` behaviour to emit `event` object containing `data` property, which matches native browser spec.
Browser no longer wraps `message` event (previously listener was being called twice - once with `event.data` and once with `data`).
Added a test for postMessage/emit behaviour

### DIFF
--- a/packages/api-tests/test/window.core.spec.js
+++ b/packages/api-tests/test/window.core.spec.js
@@ -709,7 +709,7 @@ describe('WindowCore API', function(done) {
           var eventName = `evt_${event}_count`;
           window[eventName] = [];
           var currentWin = ssf.Window.getCurrentWindow();
-          
+
           currentWin.addListener(event, evt => {
             window[eventName].push(evt.data);
             console.log(event);
@@ -744,8 +744,7 @@ describe('WindowCore API', function(done) {
         /* eslint-enable no-undef */
         return executeAsyncJavascript(app.client, script, event);
       };
-      
-      let childWindowId;
+
       const steps = [
         ...setupWindowSteps(windowOptions),
         () => addWindowListener('message'),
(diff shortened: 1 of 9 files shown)


packages/api-tests/test/window.core.spec.js
```diff
@@ -694,6 +694,69 @@ describe('WindowCore API', function(done) {
         return chainPromises(steps);
       });
     }
+
+    it.only('Should emit a message event in the window #ssf.Window.postMessage #ssf.Window.postMessage', function() {
```
BenLambertNcl: Remove only :wink:

BenLambertNcl: Also one of those tags should be `#ssf.WindowCore.postMessage` so it can link it to the correct class in the docs

packages/api-tests/test/window.core.spec.js
```diff
@@ -694,6 +694,69 @@ describe('WindowCore API', function(done) {
...
+        /* eslint-disable no-undef */
+        const script = (event, callback) => {
+          // Track the calls to the event listener
+          var eventName = `evt_${event}_count`;
+          window[eventName] = [];
+          var currentWin = ssf.Window.getCurrentWindow();
+
+          currentWin.addListener(event, evt => {
+            window[eventName].push(evt.data);
+            console.log(event);
```
BenLambertNcl: Guessing we don't need this?

BenLambertNcl (commented):
Looks good :). Think there might be more typescript updates though, `api-utility` and the compatibility api will both be on 2.3 I think

ColinEberhardt (approved):

packages/api-tests/test/window.core.spec.js
andy-lee-eng: Aha! We were just guessing how that works, but yes that makes sense.

PR #262, opened 2017-07-17
Add ability to select display when creating window

BenLambertNcl:
Passing `display: window-id` will create the window on that display, with any x and y coordinates relative to that displays bounds. However, this removes some default behavior where the window would display in the center of the screen if no x or y coordinates were passed in. This has been changed to [0, 0] to keep it consistent between containers.

Also added `clean:cache` as a script in `api-tests`, which deletes `%LOCALAPPDATA%/OpenFin/cache`, which is run before the tests. It is not run before `npm run test:ci: ...` as the cache doesn't persist on CI.


packages/api-utility/src/position.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * Only necessary in Electron and OpenFin as browser does not know about
+ * more than 1 screen.
+ */
+export const getPositionFromDisplay = (displayId: string, position: ssf.Position): Promise<ssf.Position> => {
```
andy-lee-eng: Maybe this should export a `Position` or `Display` class that has a static method? It just means that the namespace for api-utility will be better organised as it grows (i.e. not be full of loads of functions)

andy-lee-eng (approved):
It looks good, but I think I'd prefer it if we still defaulted to the middle of the screen if neither x,y coords or display id have been set. Just because defaulting to the top-left sucks.

e.g. `getPositionFromDisplay` would return `undefined` in this case so we wouldn't set position (and let Browser continue to do what it did before).

ColinEberhardt (approved):
this looks like a really useful feature.

packages/api-specification/interface/window.ts
```diff
@@ -98,6 +98,10 @@ declare namespace ssf {
      * Makes the window transparent. Default is false.
      */
     transparent?: boolean;
+    /**
+     * The id of the display to show the window on.
```
ColinEberhardt: One quick question - I presume that if the `display` property is not present, the `x` and `y` properties are relative to the origin of the given display? Probably worth mentioning in the docs.

BenLambertNcl: Origin of the primary display, but yeah I will update the comment

andy-lee-eng (approved):

BenLambertNcl:
@DevAndyLee Im still working on getting the window to display in the middle, the tests are failing, saying the height is 5px off :confused:

andy-lee-eng (approved):
👍

PR #267, opened 2017-07-18
Introduce unit-testing of api-utility project

andy-lee-eng:
Used `tape` for simple config, no need for browser
Added ability to set mock globals like `location` or `ssf`

I've added `sinon` for more sophisticated mocking and `proxyquire` to enable mocking of `require`/`import` dependencies, but not needed to use them yet for the simple initial tests


andy-lee-eng:
Note: I abandoned attempts to get AppVeyor to display the test results in its "Tests" tab - it seems `tap-xunit` and AppVeyor don't agree on what an `xunit` xml file looks like.

BenLambertNcl (approved):
LGTM. :-)

ColinEberhardt (approved):
looks nice and simple to me 👍

Issue #272, opened 2017-07-19
Browser test results show in docs when no tests have been run


JDawes-ScottLogic:
For example, `Window` tests are not run for the browser, but test results show the browser as failing:

![image](https://user-images.githubusercontent.com/11292375/28360931-e670f846-6c6e-11e7-8ae0-7ea229ca52cc.png)

BenLambertNcl:
@JDawes-ScottLogic Can this be closed, or is there more work to be done with this?

JDawes-ScottLogic:
It can be closed. There was talk about using N/A for not run, etc. but I'll log an issue with the full details for a proper discussion.

PR #273, opened 2017-07-19
Hide not run browser tests from docs

JDawes-ScottLogic:
Tests that are not run for the browser are included in the documentation
for a method, showing as 0/x. This change is to only include tests for
the browser when we have actually run some tests.

Linked to #272


BenLambertNcl (approved):
Think that will look a lot nicer! Thanks!

andy-lee-eng (approved):

PR #274, opened 2017-07-19
Implemented utility unit tests for Emitter and Display

andy-lee-eng:
Also changed the config to use the same method of typescript compile as the normal build, instead of relying on ts-node


BenLambertNcl (approved):
Nice. Looks like you found a small bug with `removeAllListeners` too!

PR #276, opened 2017-07-20
Fix OpenFin preload and remove api-bundle

### LINKED ISSUE #265: ssf-openfin preload change
Using the `ssf-openfin` cli, the `containerjs-api.js` file needs to be given in the `app.json` file, which needs to be hosted via a server. This is unintuative as the user shouldn't need to worry about the api file when using the `ssf-openfin` cli. I think we could point openfin at the preload from our cli.

## Current:
structure:
```
- app.json
- src      
  - containerjs-api.js
  - index.html
```
app.json:
```json
{
  "name": "Hello World",
  "url": "http://localhost:3000/",
  "uuid": "hello-world",
  "autoShow": true,
  "preload": "http://localhost:3000/containerjs-api.js"
}
```

This preload option in app.json shouldn't be necessary, as we preload the ssf api, not user code

## Proposed:
structure:
```
- app.json
- src      
  - index.html
```
app.json:
```json
{
  "name": "Hello World",
  "url": "http://localhost:3000/",
  "uuid": "hello-world",
  "autoShow": true
}
```

Where the openfin-cli will point at the preload script inside its build folder.

BenLambertNcl:
Fixes the `ssf-openfin` cli so that it no longer requires the preload url in the `app.json` file. Closes #265.

Also removes the `api-bundle` package, as this is not longer needed with both Electron on OpenFin using preload scripts. Documentation has been updated to reflect this. Not sure if we need to unpublish the package from npm?

@JDawes-ScottLogic I can't seem to add you as a reviewer yet, so would you mind reviewing if you have a chance please :smile:.


packages/api-browser/bundle.js
```diff
@@ -0,0 +1,13 @@
+// Need to import using this syntax as require doesn't work for local files + rollup
+import * as browserApi from './build/es/index.js';
```
andy-lee-eng: Despite the comment, I still don't quite understand this bit.

This is a JavaScript file importing the built output from TypeScript, but why can't it be `bundle.ts` and import `index.ts` directly? i.e.
```javascript
import * as browserApi from './index';
```

e.g. Does `tsconfig.json` need to be changed to include `./bundle.ts` instead of `./index.ts`? Or does this file need to be JavaScript (i.e. TypeScript doesn't like us messing about with globals)?

packages/api-demo/package.json
```diff
@@ -6,7 +6,7 @@
   "private": true,
   "scripts": {
     "serve:no-browser": "npm run serve -- --no-browser",
-    "serve": "live-server --mount=/:src --mount=/scripts/containerjs-api-bundle:node_modules/containerjs-api-bundle/build/containerjs-bundle.js --mount=/resources/notification.html:node_modules/containerjs-api-bundle/build/notification.html",
+    "serve": "live-server --mount=/:src --mount=/scripts/containerjs-api:node_modules/containerjs-api-browser/build/dist/containerjs-api.js --mount=/resources/notification.html:node_modules/containerjs-api-bundle/build/notification.html",
```
andy-lee-eng: I think we can remove the second part of the line that mounts the `notification.html` file (which did come from the bundle but won't any more).

andy-lee-eng (approved):
Looks pretty comprehensive 👍

packages/api-demo/package.json
BenLambertNcl: Ah I missed that with it being off the side of my screen :laughing:. I think we still need it though (as OpenFin still needs it, even though we preload the JS). But it does need changing from bundle to openfin.

packages/api-browser/bundle.js
BenLambertNcl: Yeah you're right, I never thought of that, was trying to reproduce the bundle in this packages, but I don't think it's necessary :-)

packages/api-demo/package.json
andy-lee-eng: I wondered about that, but then I noticed you'd added an option to ssf-openfin to generate a notification template, so I thought you'd decided to do it that way instead. But yes, I guess we will still need to map a path to wherever that file gets generated...

BenLambertNcl: That option was to make copying/creating a `notification.html` file easier, as it copies our notification into a directory. Unfortunately, this notification still needs to be served as it did before.

JDawes-ScottLogic:
@BenLambertNcl Looks good to me with my limited knowledge. :+1:

BenLambertNcl:
@DevAndyLee Updated. :-).

packages/api-browser/rollup.config.js
```diff
@@ -1,3 +1,24 @@
...
+//       jsnext: true
+//     }),
+//     commonjs({
+//       include: '**'
+//     })
+//   ],
+//   dest: 'build/dist/containerjs-api.js'
+// };
+
 import resolve from 'rollup-plugin-node-resolve';
```
andy-lee-eng: Commented section left in here.

Does this already bundle api-utility as it is?

docs/index.md
```diff
@@ -25,7 +25,7 @@ From within an empty folder, add a minimal HTML file called `index.html`:
     ContainerJS status: <span id="status">initialising</span>
   </div>
 
-  <script src="https://unpkg.com/containerjs-api-bundle@0.0.2/build/containerjs-bundle.js"></script>
+  <script src="/scripts/containerjs-api.js"></script>
```
ColinEberhardt: I'd prefer to leave this as an unpkg.com link - this example is for first-time users, so should be as simple as possible!

docs/index.md
```diff
@@ -37,7 +37,7 @@ From within an empty folder, add a minimal HTML file called `index.html`:
 </body>
 ```
 
-The above displays a simple welcome message and indicates the status of the ContainerJS APIs. The `containerjs-bundle.js` is the client-side bundle that detects the container that the page is running within (e.g. Electron, OpenFin) and provides the required API implementation. In the above code, this bundle is being downloaded from the `unpkg` CDN for simplicity.
+The above displays a simple welcome message and indicates the status of the ContainerJS APIs. The `containerjs-api.js` is the browser api, that will create the ContainerJS API if no API currently exists.
```
ColinEberhardt: API - capitals ;-)

packages/api-openfin/bin/ssf-openfin
```diff
@@ -1,19 +1,22 @@
...
 program
-  .option('-c, --config [filename]', 'ContainerJS config file', 'config.json')
-  .option('-o, --output-config [filename]', 'Where to output the OpenFin config file', 'app.json')
-  .option('-v, --version [filename]', 'Version of the OpenFin runtime to use, default is stable', 'stable')
-  .option('-u, --url [filename]', 'Url to read the new app.json file from to start OpenFin')
+  .option('-c, --config [filename]', 'ContainerJS config file', 'app.json')
+  .option('-o, --output-config [filename]', 'Where to output the OpenFin config file', 'openfinApp.json')
+  .option('-v, --version [version]', 'Version of the OpenFin runtime to use, default is stable', 'stable')
+  .option('-u, --url [url]', '(Optional) Url to read the new app.json file from to start OpenFin')
+  .option('-n, --notification [directory]', '(Optional) Generate an example notification file in the specified directory')
```
ColinEberhardt, line 13: that's a neat idea :+1:

ColinEberhardt (approved):
A few minor comments, but otherwise LGTM!

docs/index.md
BenLambertNcl: Yeah me too, but it would be out of date until we publish to npm again. Unless you think we should do that immediately after this PR?

ColinEberhardt: Yeah, I'd just do a release. As we are in 0.0.x-land at the moment, just push a release out whenever you like!

BenLambertNcl: Ok, ill update this to use the unpkg.com link and ill make a release :)

andy-lee-eng (approved):

BenLambertNcl:
Fixed, will make a release once this has been merged

packages/api-browser/rollup.config.js
BenLambertNcl: Oops. Good spot! Not sure what you mean by bundle api-utility? The node resolver includes any packages installed in node_modules, so that would include api-utility. Or have I mis-understood? :-)

andy-lee-eng: Yep, that's what I meant. It's `rollup-plugin-node-resolve` that does this isn't it?

`api-browser` already has that, but `api-bundle` didn't (so I had to add it when I created `api-utility`). You've added it to `api-openfin` already, so I think it's all good 👍

BenLambertNcl: Yeah :+1:

PR #291, opened 2017-07-20
Fix error checking in docs deployment

### LINKED ISSUE #278: Docs deployment script reports failed message on success
The documentation deployment script writes a warning on a success, as the incorrect check is performed on the `$error` variable.

Instead of
```powershell
IF ($error) {
    Write-Warning -Message 'Deploy Failed'
  }
```

it should be
```powershell
IF ($error -eq "0") {
    Write-Warning -Message 'Deploy Failed'
  }
```

JDawes-ScottLogic:
Perform the correct check for an error in the documentation deployment
script.

Fixes #278

### DIFF
--- a/deploy-script.ps1
+++ b/deploy-script.ps1
@@ -8,7 +8,7 @@ IF ($env:APPVEYOR_REPO_BRANCH -eq "master" -And (-Not (Test-Path Env:\APPVEYOR_P
   git commit -m "Update gh-pages: $env:APPVEYOR_REPO_COMMIT_MESSAGE"
   git subtree split --prefix docs -b gh-pages
   git push -f origin gh-pages:gh-pages
-  IF ($error) {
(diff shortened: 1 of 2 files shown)


BenLambertNcl (changes requested):
I have just realized I got it wrong in the gitter chat. I think it should be something like this:
```
  IF ($LASTEXITCODE -ne "0") {
    Write-Warning -Message 'Deploy Failed'
  }
```
As after reading into it a little bit, $error is an object containing an array of errors (so after the first error, will always return true), where as this uses the exit code of the last command.

It also needs changing inside `test-script.ps1`, as that will have the same problem

JDawes-ScottLogic:
I probably shouldn't have assumed that a half-remembered fix in gitter was 100% correct...

BenLambertNcl (approved):
Thanks :smile:

PR #292, opened 2017-07-20
Removed unnecessary build commands from all packages

andy-lee-eng:
Also fixed api-symphony-compatibilty project and demo to not use bundle anymore, and use ssf-openfin

Ref: https://github.com/symphonyoss/ContainerJS/issues/259


BenLambertNcl (approved):
LGTM! :+1:

PR #294, opened 2017-07-21
A couple of Browser window fixes

andy-lee-eng:
Should remove child windows from list when they are closed
Browser window should open with a default size if no width/height is set (previously a tiny window)
Added tests for closed child windows and new window sizes


JDawes-ScottLogic:
:+1:

JDawes-ScottLogic (approved):

PR #297, opened 2017-07-24
Can send message to "*" (all windows) in OpenFin and Electron

andy-lee-eng:
Browser implementation coming next
Electron needs to send the message out to all Exectron windows if windowId is *
OpenFin implemented "*" already but didn't exclude receiving the message in the same window


JDawes-ScottLogic (approved):
:+1:

PR #299, opened 2017-07-25
Implemented message routing in browser

andy-lee-eng:
All windows pass all messages received to parent and all child windows, unless the message is one they sent or have already forwarded. This allows communication between grand-child windows and siblings etc.

Guard OpenFin `app.ready` so that the initialisation doesn't get re-run.


JDawes-ScottLogic (approved):
:+1:

PR #300, opened 2017-07-25
Delete .clabot

ColinEberhardt:
Hi @maoo just wanted to double-check, we no longer need this file?


maoo:
Looking at https://github.com/symphonyoss/ContainerJS/pull/273 , it seems - from the text message - that this file is not currently used, so it should be safe to delete.

I double checked our `clabot-config` project and all the github IDs defined in this file are whitelisted.

JDawes-ScottLogic (approved):
:+1:

andy-lee-eng (approved):

Issue #301, opened 2017-07-26
Events not being emitted for all `Window` calls


JDawes-ScottLogic:
The following methods do not emit events when they are called:
* Electron
   - Capture

* Electron
   - Capture

* Browser
   - Capture

PR #307, opened 2017-07-28
Move screen snippet to window

### LINKED ISSUE #275: Move ScreenSnippet functionality to WindowCore
From the GoogleDoc:
>ScreenSnippet: Perhaps this should just be a method on a window?

This currently takes a snippet of the _window_ not the screen. Given that this is all it is doing as well, this makes more sense to be a method on a window. It also has the side effect of lowering it's visibility, i.e. it will no longer be a top level concept.

JDawes-ScottLogic:
Remove `ScreenSnippet` class and implement on `WindowCore`.
Change implementations to capture the window referred to, rather than the calling window.

Resolves #275


packages/api-browser/src/window.ts
```diff
@@ -214,5 +222,6 @@ const eventMap = {
   'focus': 'focus',
   'hide': 'hidden',
   'message': 'message',
-  'show': 'load'
+  'show': 'load',
+  'capture': 'capture'
```
andy-lee-eng: I think we probably don't need this - these are mappings for window event names.

packages/api-demo/src/window/window-api-demo.js
```diff
@@ -63,6 +63,10 @@ appReady.then(() => {
       win.addListener('close', () => {
         addListItem('close');
       });
+
+      win.addListener('capture', () => {
+        addListItem('capture');
```
andy-lee-eng: Do we fire a Window event when capture happens?

JDawes-ScottLogic: We don't currently. I'll add it to #301 .

packages/api-browser/src/window.ts
JDawes-ScottLogic: Agreed.

andy-lee-eng (approved):

PR #308, opened 2017-07-28
Added resize event to browser container

andy-lee-eng:
This is used by the symphony shim demo.
It also uses the move event, but browser won't support that.


JDawes-ScottLogic (approved):
:+1:

PR #309, opened 2017-07-28
Moved symphony demo into its own project

andy-lee-eng:
This allows us to remove the dependencies to containerjs-api-browser/electron/openfin,
which is necessary because those other projects will now need to depend on
containerjs-api-compatibility in order to bundle the code together


JDawes-ScottLogic (approved):
:+1:

PR #316, opened 2017-07-31
Remove @types/electron and update

### LINKED ISSUE #304: @types/electron-notify@0.1.4 is a breaking change
Error is:
`Namespace 'Electron' has no exported member 'BrowserWindowConstructorOptions'.`

Temporary fixed at `@0.1.3`, but will need to be updated.

JDawes-ScottLogic:
Remove @types/electron as a dependency, as it has been deprecated due to
electron shipping types alongside it's main npm package.

Update @types/notify-electron to use the latest version.

Resolves #304 and #191 .

Currently fails due to https://github.com/electron/electron/issues/9626.

### DIFF
--- a/packages/api-electron/package.json
+++ b/packages/api-electron/package.json
@@ -19,12 +19,12 @@
   },
   "homepage": "https://symphonyoss.github.io/ContainerJS/",
   "dependencies": {
-    "@types/electron-notify": "^0.1.3",
+    "@types/electron-notify": "^0.1.5",
     "commander": "^2.11.0",
     "containerjs-api-compatibility": "^0.0.6",
     "containerjs-api-specification": "^0.0.7",
     "containerjs-api-utility": "^0.0.5",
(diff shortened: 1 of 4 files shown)


JDawes-ScottLogic:
Fix has landed in `master` for this (see [here](https://github.com/electron/electron/commit/18e7354b4a19280fcd87fcc65b49f7ad31858750)), so just need to wait for the next release. This commit reverts back to using `^`, so once there is a release this PR should just need a rebuild triggering.

JDawes-ScottLogic:
beta version of electron 1.8.0 containing the fix hasn't been published to npm. Worth noting that using the caret (`^`) may not be appropriate, due to electron not following semver: https://electron.atom.io/docs/tutorial/electron-versioning/

andy-lee-eng (approved):
Do these changes still make sense (and work?). If so, then it looks ok.

PR #320, opened 2017-08-01
Symphony compatibility API updates

ColinEberhardt:
With a few minor tweaks to the compatibility API, I've been able to get symphony running in both OpenFin:

image

And electron:

image

In each case, the ContainerJS CLI was used, e.g. 

```
$ ssf-electron -c ./src/app.json --symphony 
```

or

```
$ ssf-openfin -c ./src/app.json --symphony 
```

Where the local app manifest (`app.json`) looks like the following:

```
{
    "url": "https://my.symphony.com",
    "name": "SSF Desktop Wrapper",
    "uuid": "containerjs",
    "autoShow": true
}
```

A few notes about what works and what doesn't work ... just for the record:

- The Symphony UI requires the APIs to be exposed via a `SYM_API` namespace
- Same observations as Nick Kolba - a number of un-documented APIs, e.g. `registerLogger` and `Notification.permission`
- There are a few parts of the Symphony UI code that cause failures, e.g.
 - If you try to configure desktop alert settings, it fails on `window.appbridge.ShowAlertSettings`
 - There is code that only permits certain operations under windows 
`isWrapper.electron() && isWindowsOS() ? this.Notf = appBridgeElectron.getNotificationApi() : this.Notf = window.Notification` - this should really be based on capabilities.

What works and what doesn't:

Within electron - the following features work:
- pop-out chat window
- notifications (basic title / body)
- application badge count

what doesn't work
- cannot test screen-snippets, looks like this is not implemented in Symphony UI yet
- within OpenFin notifications don't work (we'd need to locally host a `notification.html` page)
- within OpenFin pop-out windows don't work, most likely due to differences in `window.open` behaviour


packages/api-symphony-compatibility/index.ts
```diff
@@ -5,6 +5,10 @@ const api = (ssf && ssf.registerBoundsChange)
     ? ssf
     : map.ssf;
 
+// The current Symphony UI reuqires that the API is present within the SYM_API namespace, although SymphonyElecrtong supports both
```
JDawes-ScottLogic: "SymphonyElecrtong" should be "SymphonyElectron"

JDawes-ScottLogic (approved):
:+1:

packages/api-symphony-compatibility/index.ts
ColinEberhardt: Good spot - fixed.

lneir:
@ColinEberhardt in next release 1.47 we have move to window.ssf

i'm having a conversation today with Nick and we'll fix the registerLogger issue - that is private api and so of course shouldn't be dependent there - we'll get a fix in next release.

Notification.permission is implemented to provide compatibility with browser notifications - i'll update documentation to note that this is needed.

showAlertSettings will be available in next release of front-end code.

for windows... our front-end code uses the custom notifications we have implemented in symphony electron but for mac our product team wants to use the electron's built in native notifications (which in turn uses mac native os notifications).

PR #323, opened 2017-08-03
Fix callbacks in getCurrentWindows

### LINKED ISSUE #314: getCurrentWindow inconsistent callback behavior
`getCurrentWindow(callback, errorCallback)` always returns a `Window` but it may not be ready to use until the `callback` has been called (observed in OpenFin implementation).

However, sometimes the callback is never called; as a consumer of the API, I don't know when I can safely use the `Window` object returned by this method.

Annotated code in the API with comments to highlight the issue:

```ts
static getCurrentWindow(callback?: (win: Window) => void, errorCallback?: (err?: any) => void) {
  if (currentWindow) {
    // In this branch, callback will never be called,
    // how do I know that currentWindow is ready to use?
    return currentWindow;
  }
  
  // In this branch, I have to wait until callback is called,
  // otherwise currentWindow may not be ready
  currentWindow = new Window(null, callback, errorCallback);
  return currentWindow;
}
```

BenLambertNcl:
Fixes #314.


packages/api-openfin/src/window.ts
```diff
@@ -192,14 +192,16 @@ export class Window extends Emitter implements ssf.Window {
           this.innerWindow = app.getWindow();
           this.id = `${this.innerWindow.uuid}:${this.innerWindow.name}`;
 
-          fin.desktop.InterApplicationBus.publish('ssf-new-window', {
-            windowName: this.innerWindow.uuid,
-            parentName: options.child ? Window.getCurrentWindow().innerWindow.uuid : null
+          Window.getCurrentWindow((win) => {
```
JDawes-ScottLogic: Does this change introduce some potential latency on publishing the new window event? Previously we'd only be waiting on the current window if we were creating a child window, while now we always need the callback to be fired.
Would it make sense to grab it from the `fin.desktop.Window.getCurrent()` call we're making above to get the window options?

packages/api-symphony-compatibility/src/mapping.ts
```diff
@@ -29,14 +29,18 @@ export namespace map {
     /** API defined at https://symphonyoss.atlassian.net/wiki/display/WGDWAPI/RegisterBoundsChange+API */
     static registerBoundsChange(callback: Function) {
       // Not fully implemented in containerjs, only listens to the current window, not its children as well
-      containerjsSsf.Window.getCurrentWindow().addListener('resize', () => {
-        containerjsSsf.Window.getCurrentWindow().getBounds().then((bounds) => {
-          callback(bounds);
+      containerjsSsf.Window.getCurrentWindow(win => {
```
JDawes-ScottLogic, line 32: Can this now be done in a single callback?

packages/api-tests/test/window.core.spec.js
```diff
@@ -681,6 +681,23 @@ describe('WindowCore API', function(done) {
...
+      const getCurrentWindow = () => {
+        const script = (callback) => {
+          ssf.Window.getCurrentWindow(() => callback(true));
+        };
+        return executeAsyncJavascript(app.client, script);
+      };
+
+      const steps = [
+        () => getCurrentWindow(),
+        () => getCurrentWindow(), // Make sure the callback is called twice
```
JDawes-ScottLogic, line 694: Why do we need to make sure it's called twice?

JDawes-ScottLogic (approved):
:+1:

packages/api-tests/test/window.core.spec.js
BenLambertNcl: The problem happened when get current window was called a second time, as it was being called correctly when the window was initially created, but not when retrieving the cached version

packages/api-openfin/src/window.ts
BenLambertNcl: Not sure what you mean by this :D. There will be a very slight delay sending the `ssf-new-window` event to the hidden window. `fin.desktop.Window.getCurrent()` returns the OpenFin specific window, whereas `Window.getCurrentWindow()` returns the ContainerJS window object (or passes it to the callback :) ) So I don't think we could use the OpenFin one.

BenLambertNcl:
Updated. :)

JDawes-ScottLogic (approved):
:+1:

Issue #338, opened 2017-08-09
Revisit the UX for test results in the docs


JDawes-ScottLogic:
The test results in the documentation is a nice feature, however it's not particularly clear that they are test results.

- Add "test results" text?
- Group them into collapsible element perhaps?

PR #348, opened 2017-09-26
Show combined test result under label "Tests"

andy-lee-eng:
Click to reveal detailed test report for method/property

This is an attempt to improve the UX as mentioned in issue #338


andy-lee-eng:
subscribe-test-closed

andy-lee-eng:
subscribe-test-open

packages/api-specification/transform-type-info.js
```diff
@@ -62,8 +62,15 @@ if (program.testfile) {
     const testReport = JSON.parse(fs.readFileSync(program.testfile));
     Object.keys(testReport).forEach((testMethod) => {
       const getSelectionString = (type, name, childName) => `..{(.kindString === "Class" || .kindString === "Interface") && .name === "${name}"}.children{.kindString === "${type}"` + (childName ? `&& .name === "${childName}"}` : '}');
+      const getPassed = obj => obj ? obj.passed : 0;
+      const getTotal = obj => obj ? obj.total : 0;
```
JDawes-ScottLogic, line 66: It looks like this may introduce a variation on #273 if we don't have a set of results? Do we want to be showing an empty test run in this case?

andy-lee-eng: Do you mean where Browser doesn't implement the feature, like this?

getbounds-test-open

andy-lee-eng: I think the problem wouldn't occur because it only calculates the `combined` score if there's a `results` object in the `testReport` for that method.

There are a couple of odd ones where it reports `Electron 0/0 OpenFin 0/0`, but that was already happening before (e.g. Window.reload). The test-report.json file actually has that data in it.

JDawes-ScottLogic: That's exactly the problem, but you're right that it's now removed from the test results before this runs. :+1:

JDawes-ScottLogic (approved):

JDawes-ScottLogic:
:+1:

Issue #281, opened 2017-07-20
Ensure documentation is clear that browser doesn't implement Window


JDawes-ScottLogic:
From the GoogleDoc:
>Make it very clear that browser doesn’t implement ‘Window’ but instead implements ‘WindowCore’

PR #351, opened 2017-09-27
Docs explain that Window is not available in Browser

andy-lee-eng:
Window refers to WindowCore and vice-versa for the relevant containers.
Improved the rendering of commends so that breaking up a comment with a double new-line works (previously it was merged into a single line).


JDawes-ScottLogic (approved):
:+1:

andy-lee-eng:
Fixes Issue #281

Issue #347, opened 2017-09-14
Automate the release process


JDawes-ScottLogic:
The release process needs to be automated, as both @ColinEberhardt and myself have failed to release a correct version at the first time of asking. This should be a non-issue, as the CI tooling should be able to release, removing the risks associated with manually releasing.

With this in mind, the following options seem to be available:

### Release Branch:
Have a dedicated branch for releasing from. Update the version, then merge to the branch, allowing CI to detect this and trigger a release.

Pros:
- Simple to setup
- Works with existing process (`lerna` independent versioning)

Cons:
- Requires manual intervention to release.
- Releases only contain a subset of the code on `master`.
- Documentation is therefore out of date.

### Full CI release:
Setup CI to release whenever a merge is made to `master`. 

Pros:
- Release process is fully automated.
- Fixes are released as soon as they are approved to `master`.
- Documentation is fully up to date.

Cons:
- Requires a lot more setup and configuration i.e. `commitizen`, `sematic-release`.

The manual step with a release branch is the logical first step to automating this, as it takes the process out of the developers hands and makes it reproducible. However, in the long term it makes sense to remove all developer involvement in the releases and to let the CI handle the release. Discussions about this have started in #232 .

PR #358, opened 2017-09-29
Add script for publishing packages

JDawes-ScottLogic:
Script for releasing via Appveyor.

- Will only trigger from `release` branch`.
- Requires `npm_release_token` environment variable set.
- Assumes that a build has taken place already.
- Performs `npm publish` for all non-private packages it can find under `packages\`.

Associated with #347


andy-lee-eng (approved):
Looks good (at least, as far as I can tell)!

andy-lee-eng:
I think we've decided to stick with the documented manual release process for now (i.e. using Lerna), on the basis that it is sufficient for a small team with occasional releases.

Issue #327, opened 2017-08-03
Improve the CLI interfaces


ColinEberhardt:
A few notes on how the `ssf-electron`, `ssf-openfin` interfaces could be improved ...

 - currently `ssf-electron` fails with an error if no config file is passed
 - they should have an option for reporting their version number
 - the README for each (as displayed in npm https://www.npmjs.com/package/containerjs-api-openfin), should be focused on their use as a CLI tool. Perhaps just document them by showing the output of `ssf-openfin --help` in the readme to avoid repeating things?
 - should be able to fetch the app.json via HTTP as well as file://
 - you should be able to specify whether to show / display the developers tools (context menu / system menu)

PR #359, opened 2017-09-29
Can download app.json file from http, and OpenFin developer menu

andy-lee-eng:
Issue #327 

Electron already has a -d/--developer option for whether to show the Developer menu. Added a similar option for allowing the OpenFin context menu.

Also added the ability to download the app.json file from http or https

### DIFF
--- a/packages/api-electron/main.js
+++ b/packages/api-electron/main.js
@@ -5,6 +5,8 @@ const path = require('path');
 const program = require('commander');
 const download = require('download');
 const packageJson = require('./package.json');
+const promisify = require('util.promisify');
+const readFileAsync = promisify(fs.readFile);
 
 program
   .version(packageJson.version)
@@ -60,20 +62,12 @@ function readConfigFile() {
   if (filePath.toLowerCase().startsWith('http://') || filePath.toLowerCase().startsWith('https://')) {
     return download(filePath);
   } else {
-    return new Promise((resolve, reject) => {
-      const configFile = path.join(process.cwd(), program.config);
-      if (fs.existsSync(configFile)) {
-        fs.readFile(configFile, (err, data) => {
-          if (err) {
-            reject(err);
-          } else {
-            resolve(data);
-          }
-        });
-      } else {
-        reject(new Error(`Config file ${configFile} does not exist`));
-      }
-    });
+    const configFile = path.join(process.cwd(), program.config);
+    if (fs.existsSync(configFile)) {
+      return readFileAsync(configFile);
+    } else {
+      return Promise.reject(new Error(`Config file ${configFile} does not exist`));
+    }
   }
 }
 
--- a/packages/api-electron/main.js
+++ b/packages/api-electron/main.js
@@ -1,7 +1,9 @@
 const { app } = require('electron');
 const ssfElectron = require('./index.js');
 const fs = require('fs');
+const path = require('path');
 const program = require('commander');
+const download = require('download');
 const packageJson = require('./package.json');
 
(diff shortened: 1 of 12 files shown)


andy-lee-eng:
The last commit fixes two failing tests that were broken when the -d option was originally added to Electron, which happened in this commit: https://github.com/symphonyoss/ContainerJS/commit/bdeab1f634c013fde189d163dc1eec54bf9d2217

packages/api-electron/main.js
```diff
@@ -17,49 +19,64 @@ program
...
 function loadConfig() {
-  // the config json is passed to the ssf-electron bin script
-  const configFile = program.config ? process.cwd() + '/' + program.config : null;
-  const launchUrl = program.url;
-  if (configFile) {
-    if (fs.existsSync(configFile)) {
-      const config = JSON.parse(fs.readFileSync(configFile, 'utf8'));
+  if (program.config) {
+    return readConfigFile()
+      .then(data => new Promise((resolve, reject) => {
```
ColinEberhardt: Do you have to manually construct a Promise here?


```
   .then(data => new Promise((resolve, reject) => {
        const config = JSON.parse(data);

        if (program.url) {
          config.url = program.url;
        }

        if (config.url) {
          resolve(config);
        } else {
          reject(new Error('You must specify an URL (--url) or a config file containing an url (--config)'));
        }
      }));
```

could you instead throw rather than reject?

```
   .then(data => {
        const config = JSON.parse(data);

        if (program.url) {
          config.url = program.url;
        }

        if (config.url) {
          return config;
        } else {
          throw new Error('You must specify an URL (--url) or a config file containing an url (--config)');
        }
      });
```

packages/api-electron/main.js
```diff
@@ -1,7 +1,9 @@
 const { app } = require('electron');
 const ssfElectron = require('./index.js');
 const fs = require('fs');
+const path = require('path');
 const program = require('commander');
+const download = require('download');
```
ColinEberhardt: Are you using the `download` package as a way to get a promise-based interface for http requests?

Perhaps it is better to look for a more generic approach that can turn any node-callback interface into a promise? e.g. [util.promisify](http://2ality.com/2017/05/util-promisify.html) (or polyfill).

packages/api-electron/main.js
```diff
@@ -17,49 +19,64 @@ program
...
 
-function fileError(filename, err) {
-  consoleError('Error with file: ' + filename);
-  consoleError(err);
+function readConfigFile() {
+  const filePath = program.config;
+  if (filePath.toLowerCase().startsWith('http://') || filePath.toLowerCase().startsWith('https://')) {
+    return download(filePath);
+  } else {
+    return new Promise((resolve, reject) => {
```
ColinEberhardt: If you convert the fs.readFile API into a promise-based interface, can you remove the need to manually construct a Promise?

ColinEberhardt (changes requested):
Looks good to me  - just a few comments around the use of Promises. In most cases you shouldn't have to construct a Promise explicitly via its constructor - might be worth exploring util.promisify

packages/api-electron/main.js
andy-lee-eng: Actually it was more a way to easily fetch a file without having to worry about http/https and reading a stream etc.

Can we use Node v8 functions? I don't think we currently specify a minimum version of Node, and the Node LTS is 6.11.4.

ColinEberhardt (approved):
Thanks for the updates

PR #360, opened 2017-10-03
Test setup to handle OpenFin additional window behaviour

andy-lee-eng:
Get the window handles that are created as part of initialisation, so that they can be excluded from the tests

Also allow a small delay when closing windows, to ensure it has a chance to close any related child windows

### DIFF
--- a/packages/api-tests/test/test-helpers.js
+++ b/packages/api-tests/test/test-helpers.js
@@ -4,8 +4,13 @@ const executeAsyncJavascript = (client, script, ...args) => {
   return client.executeAsync(script, ...args);
 };
 
+// In OpenFin, an extra hidden window gets created that interferes
+// with tests that counts windows or selects windows by index. This
+// initialisation gets the added window handles, so they can be
+// excluded later
 let addedHandles = [];
 const initialiseWindows = (client) => {
+  // Script to wait for app.ready()
   const script = (callback) => {
     if (window.ssf) {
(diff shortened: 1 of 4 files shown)


packages/api-tests/test/test-helpers.js
```diff
@@ -4,9 +4,37 @@ const executeAsyncJavascript = (client, script, ...args) => {
...
+    } else {
+      callback();
+    }
+  };
+
+  // Get a list of any extra window handles that get create during initialisation
+  return executeAsyncJavascript(client, script)
+      .then(() => client.windowHandles())
+      .then(handles => {
+        addedHandles = handles.value.splice(1);
```
ColinEberhardt: Probably worth adding a comment regarding why the first window handle is removed.

packages/api-tests/test/window.core.spec.js
```diff
@@ -26,13 +27,41 @@ const setupWindowSteps = (windowOptions) => [
   () => app.client.waitForVisible('.visible-check')
 ];
 
+const closeWindowSteps = () => [
+  () => callAsyncWindowMethod('close'),
+  () => app.client.pause(100)
```
ColinEberhardt: worth adding a comment about why a pause is needed?

ColinEberhardt (approved):
other than a couple of minor comments, looks good to me.

ColinEberhardt:
Thanks for the updates :+1:

Issue #331, opened 2017-08-04
containerjs-api-specification shows old README.md content


JDawes-ScottLogic:
Looking at the npm page for containerjs-api-specification, it is still displaying the old `README.md` content, however this was deleted by 49252349b23970080b6f8bf20890ece0f31cf01a.

This could be an artifact of #328 (there are recently resolved npm issues similar to this), and may be resolved by republishing, but do we want to re-add the `README.md` so the package has a proper description on npm?

PR #363, opened 2017-10-05
Added a couple of missing README.md files

andy-lee-eng:
For non-private npm packages, just providing basic info redirecting
to ContainerJS website


andy-lee-eng:
Fixes Issue #331

Issue #346, opened 2017-08-30
Update packages to only include the necessary files


JDawes-ScottLogic:
The npm packages currently contain more information than is necessary for consuming them, e.g. tests, `tsconfig.json`. We should only be packaging those files necessary for a release.

This can be achieved either via a blacklist (`.npmignore`) or a whitelist (the `files` property of `package.json`). Given that we tend to build everything into a single output folder, the whitelist approach seems to make most sense.

JDawes-ScottLogic:
Non-private packages to clean up:

- [x] containerjs-api-browser
- [x] containerjs-api-electron
- [x] containerjs-api-openfin
- [x] containerjs-api-specification
- [x] containerjs-api-compatibility
- [x] containerjs-api-utility

PR #366, opened 2017-10-10
Ignored non-releasable files in package/api-electron

andy-lee-eng:
Issue #346

### DIFF
--- a/packages/api-electron/package.json
+++ b/packages/api-electron/package.json
@@ -19,11 +19,7 @@
   },
   "homepage": "https://symphonyoss.github.io/ContainerJS/",
   "dependencies": {
-    "@types/electron-notify": "^0.1.5",
     "commander": "^2.11.0",
-    "containerjs-api-compatibility": "^0.0.6",
-    "containerjs-api-specification": "^0.0.7",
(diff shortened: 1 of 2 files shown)


packages/api-electron/package.json
```diff
@@ -33,7 +29,11 @@
     "ssf-electron": "./bin/ssf-electron"
   },
   "devDependencies": {
+    "@types/electron-notify": "^0.1.5",
     "concat": "^1.0.3",
+    "containerjs-api-compatibility": "^0.0.6",
+    "containerjs-api-specification": "^0.0.7",
+    "containerjs-api-utility": "^0.0.5",
```
ColinEberhardt, line 36: is this really a dev dependency?

andy-lee-eng: Because it gets rolled-up into build/dist/containerjs-api.js (or containerjs-api-symphony.js), which is the preload file that contains the entire compiled ContainerJS library.

ColinEberhardt: Aha - I thought there was a good reason. Thanks for educating me :-)

ColinEberhardt (approved):

Issue #315, opened 2017-07-29
Inconsistency between promises and callbacks


jleft:
Some parts of the API use callbacks (e.g. `ssf.Window.getCurrentWindow()`), whilst others use promises.

If possible, it would be nice if there was a consistent approach, so it's more intuitive for the consumer.

PR #371, opened 2017-10-10
Change getCurrentWindow to return a promise

andy-lee-eng:
This is a breaking change, but the previous behaviour was effectively broken anyway, because the calling code needed to wait for a callback, but the callback is only called the first time.

See Issue #314


andy-lee-eng:
Also addresses issue #315

